PR #341, opened 2026-06-03
feat(Evaluation): records are processed in parallel, speeding up runs

alexisjamet:


alexisjamet:
Draft: to be discussed with @did @jdoucy @OlivierDoucy

PR #362, opened 2026-06-09
core: workers are split into two pools (CPU pool or GPU pool)

alexisjamet:


apps/api/src/common/workers-health/workers-health.controller.ts
```diff
@@ -13,7 +17,7 @@ export class WorkersHealthController {
 
   constructor(
     @InjectDataSource() private readonly dataSource: DataSource,
-    @InjectQueue(DOCUMENT_EMBEDDINGS_QUEUE_NAME) private readonly bullmqQueue: Queue,
+    @Inject(WORKERS_HEALTH_QUEUE) private readonly bullmqQueue: Queue,
```
did, line 20: `@InjectQueue` -> `@Inject`. Is that normal?

did: write a short comment about it?

alexisjamet:
closed

PR #437, opened 2026-06-19
poc: rbac / policies version 2

did:


docs/specs/scoped-rbac.md
```diff
@@ -0,0 +1,699 @@
...
+        └── ChatSession
+```
+
+Future resources can be added without changing the authorization model.
+
+---
+
+### Global Rules
+
+**Rule 1 — Role assignments apply downward**
```
jdoucy, line 119: Not fully agreed, lets discuss

jdoucy: I mean the rules should be hardcoded (in fact I'm not sure to fully understand this part)

docs/specs/scoped-rbac.md
```diff
@@ -0,0 +1,699 @@
...
+
+---
+
+### Data Model
+
+```
+roles
+- id
+- key
+- name
```
jdoucy, line 170: We can add a desc

docs/specs/scoped-rbac.md
```diff
@@ -0,0 +1,699 @@
...
+- scope_type
+
+permissions
```
jdoucy, line 173: add a name and a desc (or its in i18n maybe)

docs/specs/scoped-rbac.md
jdoucy: or its in i18n (in this case, why key is not the id?)

docs/specs/scoped-rbac.md
```diff
@@ -0,0 +1,699 @@
...
+No access to the parent project or organization.
+
+---
+
+### What This Model Does NOT Support
+
+- Negative permissions
+- ABAC (attribute-based access control)
+- Complex conditional authorization
+- Automatic collection filtering
```
jdoucy, line 265: It was one of the original promise

PR #523, opened 2026-07-13
feat: add rbac tables and organization role seeding

did:


packages/api-contracts/src/organizations/organizations.routes.ts
```diff
@@ -1,8 +1,16 @@
...
 import { defineRoute } from "../helpers"
-import type { OrganizationDto, UpdateOrganizationRequestDto } from "./organizations.dto"
+import type {
+  OrganizationDto,
+  UpdateOrganizationRequestDto,
+  UserOrganizationListItemDto,
+} from "./organizations.dto"
 
 export const OrganizationsRoutes = {
+  listOrganizations: defineRoute<ResponseData<UserOrganizationListItemDto[]>>({
```
alexisjamet: rename into getAll

packages/api-contracts/src/organizations/organizations.routes.ts
```diff
@@ -1,8 +1,16 @@
...
+    method: "get",
+    path: "organizations",
+  }),
   createOrganization: defineRoute<
```
alexisjamet: rename into createOne

packages/api-contracts/src/organizations/organizations.dto.ts
```diff
@@ -18,6 +20,20 @@ export type OrganizationDto = {
...
 }
 
+export type UserOrganizationListItemProjectDto = {
+  id: string
+  name: string
+  featureFlags: FeatureFlagsDto
+}
+
+/** Slim organization row for GET /organizations. */
+export type UserOrganizationListItemDto = {
```
alexisjamet: pourquoi `UserOrganizationListItem`? pourquoi c'est pas simplement `Organization` ?

packages/api-contracts/src/organizations/organizations.dto.ts
```diff
@@ -18,6 +20,20 @@ export type OrganizationDto = {
   projects: ProjectDto[]
 }
 
+export type UserOrganizationListItemProjectDto = {
```
alexisjamet: pourquoi ça s'appelle pas simplement Workspace our Projet?

apps/web/src/common/features/projects/projects.middleware.ts
```diff
@@ -30,7 +31,7 @@ listenerMiddleware.startListening({
       }),
     )
 
-    await listenerApi.dispatch(fetchMe())
+    await Promise.all([listenerApi.dispatch(fetchMe()), listenerApi.dispatch(fetchOrganizations())])
```
alexisjamet: pourquoi fetch les organizations quand on crée un projet?

packages/api-contracts/src/organizations/organizations.dto.ts
did: because it's not an "real" organization entity. It's a made up model without all the organization attributes + it includes permissions AND a list of projects.
It's tailored for the onboarding screen.

If at some point of the project, we need to list raw organizations, we'll modify this endpoint and the returned types.

apps/web/src/common/features/projects/projects.middleware.ts
did: same answer as why we call `fetchMe` when we create a new project.
Actually, we could get rid of the fetchMe dispatch in that case since we don't use the list of organizations from the fetchMe endpoint. Let me check.

packages/api-contracts/src/organizations/organizations.dto.ts
did: Because it's a project within the context of the onboarding screen. We don't need all the information of a project.

apps/api/src/domains/organizations/organization.model.ts
```diff
@@ -0,0 +1,12 @@
+export type OrganizationProjectModel = {
+  id: string
+  name: string
+  featureFlags: string[]
+}
+
+export type OrganizationModel = {
+  id: string
+  name: string
+  permissions: string[]
```
alexisjamet: pourquoi les permissions sont sur l'organization?

did: Ok, we don't need them at this stage but that's the solution I chose to let the frontend know about which action it can apply to the organization resource.
Example:
```json
"data": [
        {
            "id": "7074f0ac-d3a8-4ad9-af64-a04f32b0cfa9",
            "name": "Acme Corp",
            "permissions": [
                "organization.delete",
                "organization.read",
                "organization.update",
                "project.create",
                "project.read"
            ],
            }
            ]
```            
So, when you display a list of organizations, you know that the current user can:
- delete this organization
- update this org
- read all the projects of this org
- create a new project in this org.

PR #562, opened 2026-07-21
fix(api): force full-page OCR on scanned PDFs with no text layer

### LINKED ISSUE #561: PDF processing fails: Docling produces no embed_text chunks
Uploading certain PDFs as a websource/document fails during processing. Both normal and 90°-rotated PDFs are affected, and the failing files are not unusually large, so this doesn't look like a size/resource issue.

Error observed:

`Docling produced no embed_text chunks for MIME type: application/pdf`

Reproduction: Confirmed both in production and locally. Reproducing it locally (not just on one deployment) rules out an infra/environment-specific cause and points to Docling's PDF handling itself.

Workaround found: Converting the failing PDF to an image (screenshot) first, then feeding that image into Docling instead of the PDF, works — the image is processed correctly. 

issues is due to some pdf having vectorized text, since its not an image the bitmap threshold doesn't trigger and since there isn't any "selectable" text docling doesn't get any text from the page.

Noe-Bosc-Haddad:
- some scan-to-PDF pipelines draw pages as vector outlines (no text layer, no bitmap images), so docling's bitmap-coverage OCR trigger never fires and extraction returns empty output
- classify each PDF with docling's own backend (~160ms/file) and set ocr_options.force_full_page_ocr only when the text layer is missing, keeping the fast path for native-text PDFs (always forcing costs 1.5-2.9x)
- force_full_page_ocr must be set on the nested ocr_options; the PdfPipelineOptions kwarg is silently dropped by pydantic
- log the decision as force_full_page_ocr in the chunker_runtime_info event
- add a real vector-scan fixture (test-1.pdf) plus classifier, wiring, and end-to-end tests


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 - French UI: the "Add library" button on an agent's Resources tab is now translated
 - Markdown (`.md`) files can be uploaded as documents, like other plain-text formats
 - Extraction agents accept Markdown and plain-text documents, in addition to PDF, images and CSV
+- Fix some scanned PDF documents importing with no extracted text — they are now detected and fully OCR'd
```
jdoucy: @Noe-Bosc-Haddad retire  — they are now detected and fully OCR'd

PR #597, opened 2026-07-28
refactor(agents): cite knowledge base passages by ref in the sources tool

alexisjamet:
Stacked on #595 — review that one first; this PR only adds the commit on top.

## Problem

The model had to copy chunk ids, document ids, titles and source types out of the `lookup_knowledge_base` output and back into the `sources` tool. It did that unreliably: wrong ids, wrong titles, web-crawled pages silently omitted. The tool also took five nested fields for information the server already had, and every call cost an extra LLM round-trip whose only job was to read `"Sources received"`.

## What changes

**The model no longer sees any id.** `lookup_knowledge_base` returns `{ ref, documentTitle, content }` per passage; the real records live in a per-turn `RetrievedPassageRegistry` (refs keep counting across successive lookups, a chunk retrieved twice keeps its ref).

**`sources` takes one parameter**, `refs: [1, 4]`. It resolves the refs against the registry and rebuilds what the UI reads — `documentId`, `documentTitle`, `documentSourceType`, `chunks[].chunkId` and a verbatim excerpt (capped at 1200 characters). Invented refs are logged and dropped; if nothing resolves, no tool call is persisted, so no empty sources panel. The persisted shape is unchanged, so the front keeps working — only the excerpt rendering was adjusted, since it is a real paragraph now instead of a model-written one-liner.

**`sources` is declared a terminal tool.** `buildToolLoopStopConditions` ends the tool loop when a terminal tool is called *in the same step as the answer*, instead of spending a turn on the acknowledgement. Plumbed generically through `LLMConfig.terminalToolNames` → `BuiltTools.terminalToolNames` → `ToolLoopAgent.stopWhen`, sub-agents included, keeping `stepCountIs(20)` (the SDK's implicit default) explicit.

## Ordering matters for small models

Models served through vLLM (`gemma-4-26b`) emit either text or a tool call in a step, never both — a langfuse trace of a sourced answer shows 22 output tokens for the tool call and 480 for the text. So the prompt asks for the citation **before** the answer, and a citation step with no text keeps the loop running. Asking such a model to cite "alongside your answer" makes it skip the citation entirely.

`sources` returns a `nextStep` note for that order ("answer it now if you have not yet"), replacing the previous `"Say nothing in response to the user"` output, which was counter-productive when the answer had not been written yet.

## Side effects

- `sources` is no longer built when retrieval is off — it could only produce empty sources there.
- The prompt no longer needs the "do not omit web-crawled pages" plea: the server cannot omit them.


jdoucy:
duplicated #595

PR #623, opened 2026-07-31
feat(mcp): forward conversation context to MCP servers as request headers

jdoucy:
## Why

An MCP server had no reliable way to know **which agent, session or embedded visitor** a tool call came from. The only channel was a tool parameter the model had to fill from its prompt — which is neither guaranteed (the model can omit it, or emit the call in the wrong channel entirely) nor trustworthy (a copied string can be mangled or invented).

Concrete need: one MCP server serving two agents, a test one and a production one, has to tell them apart, and an alert it forwards has to carry the visitor identifier the embedding page provided. Both must be plumbing, not prompt engineering.

## What

**Conversation context, on every MCP call, for every server.** Built in code from the streaming scope and sent as headers on the MCP transport:

| Header | Value |
|---|---|
| `X-Bayes-Agent-Id` | the calling agent |
| `X-Bayes-Session-Id` | the conversation or embed session |
| `X-Bayes-External-Visitor-Id` | the identifier the embedding page attached to a public session (absent on internal sessions) |

The public session proxy now carries that visitor identifier so it can reach the transport. The embed controller already received and stored it, nothing changes there.

Context headers are applied after the server's own configuration, so a configuration cannot spoof them.

## What consumes it, and how a server tells its deployments apart

The MCP server decides, from the headers alone:

- **`X-Bayes-External-Visitor-Id` present** means an embedded session, so a real user is behind the call. Absent means playground, evaluation or app: the server can record the call without notifying anyone.
- **Environment** comes from the agent id. The server keeps the list of agent identifiers it considers production in its own configuration and matches `X-Bayes-Agent-Id` against it. Nothing to declare on the platform side, nothing a caller can claim, and an unlisted agent falls back to the test label rather than the other way round.

That resolution belongs to the server, not here: the platform sends facts, the server decides what they mean for it.

## Privacy

These headers reach third-party servers, so the set is deliberately minimal: no user name, no message content, no organization or project identifier. Adding a field means deciding it is acceptable for every MCP server the platform can be pointed at.

## Also in this PR

`McpServerConfig` gains an optional `headers` record, sent on every call to that server. It is a generic capability, no longer needed by the case above now that the environment comes from the agent id. It costs no migration (the config is an encrypted blob) and the Studio dialog does not expose it. Say the word and I drop it to keep the PR minimal.

## Tests

Header builder: context forwarded, visitor header omitted when absent, server auth and static headers preserved, static headers cannot override the context, empty names and values dropped (they would make the transport throw), nothing sent when there is nothing to send. Plus an e2e assertion that the context reaches `connect` through the real pipeline.

## Verified against a real server

The first consumer is deployed and tested end to end. Without the visitor header the call is recorded on the server and notifies nobody; with it, the notification goes out and carries the identifier. The tool that receives it no longer declares the identifier as a parameter at all, so it can only come from the transport.

`trivy` is red on expired ignore entries in `.trivyignore.yaml` (30/06), unrelated to this change.

### DIFF
--- a/apps/api/src/domains/agents/shared/agent-session-messages/streaming/streaming-session.types.ts
+++ b/apps/api/src/domains/agents/shared/agent-session-messages/streaming/streaming-session.types.ts
@@ -15,6 +15,11 @@ export type PublicStreamingSessionProxy = {
   id: string
   traceId: string
   organizationId: string
+  /**
+   * Identifier the embedding page attached to this session (for France
+   * Travail: the "identifiant DE"). Forwarded to MCP servers as context.
+   */
+  externalVisitorId?: string | null
   messages: AgentMessage[]
 }
 
--- a/apps/api/src/domains/agents/shared/agent-session-messages/streaming/streaming.service.ts
+++ b/apps/api/src/domains/agents/shared/agent-session-messages/streaming/streaming.service.ts
@@ -155,6 +155,7 @@ export class StreamingService extends ServiceWithLLM {
     notifyClient,
     sessionState,
     sessionResult,
+    externalVisitorId,
   }: {
     connectScope: RequiredConnectScope
     publicSessionId: string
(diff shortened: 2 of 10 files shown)


jdoucy:
Closing: the MCP no longer needs forwarded context. The alert MCP will log server-side instead of notifying a channel, so it needs neither the visitor identifier nor a deployment tag. Reopen if a server ever needs deterministic caller context — the design and its tests are in this branch's history.

jdoucy:
Reopening with a narrower scope: context headers only for EMBED sessions. Off-embed (playground, evaluations, app) the platform sends nothing and the alert MCP logs instead of notifying a channel.

Issue #25, opened 2026-04-02
Agent Builder UI/UX (v2)


jdoucy:
Health indicators and Evaluation scores integrated into Agent cards.

---
_Migrated from bayesimpact/DEPRECATED-agent-studio#88_

Issue #23, opened 2026-04-02
Classifier-Based Guards


jdoucy:
Dedicated safety models (Llama-Guard/BERT) for pre/post-processing.

---
_Migrated from bayesimpact/DEPRECATED-agent-studio#86_

Issue #16, opened 2026-04-02
Evaluation Runner v2


jdoucy:
Automated bulk testing suite across datasets and all agents (also conv agents).

alexisjamet:
closed in #563

Issue #13, opened 2026-04-02
RAG v2


jdoucy:
v2 of RAG related features based on first feedbacks

---
_Migrated from bayesimpact/DEPRECATED-agent-studio#74_

PR #632, opened 2026-08-07
feat: docling/playwright crawler engine (alternative to spider.cloud)

Mascode-Dev:


alexisjamet:
## Code review — Docling/Playwright crawler engine

Reviewed at max depth: ten finder passes over the diff plus a 13-agent adversarial verification pass, each finding traced end-to-end against the PR head. **All 19 verified claims were confirmed; none refuted.** Findings are grouped by severity, each with a suggested fix.

The overarching theme: as written, the engine **cannot succeed in a deployed environment**, and where it does run it tends to fail **silently** (documents marked "completed"/"pending" rather than "failed"). Most of the operational gaps trace back to one root cause — the engine was added as a *parallel vertical* (a full copy of the url-crawling stack) rather than as a strategy behind the existing crawl pipeline, so cross-cutting behaviour (cancel routing, recrawl, metrics) silently diverged.

---

### 🔴 Blockers — the engine can't work once deployed

1. **No runtime image installs Playwright Chromium** — `apps/api/Dockerfile`
   `playwright@1.62.0` ships no browser-download install script, and no build stage runs `npx playwright install chromium` or installs Chromium's OS libs (libnss3/libatk/libgbm/libasound2…). `npm ci` downloads no browser. Result: `chromium.launch()` throws *"Executable doesn't exist"* in every deployed container and every docling crawl job fails.
   **Fix:** In the cpu/gpu worker runtime stages, add `RUN npx playwright install --with-deps chromium` (needs network + the playwright CLI at build time), and ensure the browser cache lands in the final image — either keep the default `~/.cache/ms-playwright` in the runtime stage or set `PLAYWRIGHT_BROWSERS_PATH=0` so browsers co-locate with `node_modules` and get copied with it. Add a CI/smoke check that actually calls `chromium.launch()` so a missing browser fails the build, not production.

2. **`WORKER_QUEUE_NAMES` templates never list `docling-crawling`** — `apps/api/.env-example:75`, `infra/docker-compose.api-workers-smoke.yaml:94` & `:129`, `apps/api/jest.setup-early.ts:9`
   `docling-crawling` was added to `KNOWN_WORKER_QUEUE_NAMES`/the module registry but not to any allowlist template. `workers-app.module.ts` only loads `DoclingCrawlingWorkersModule` when the queue is listed (unset/empty *throws*, it doesn't default to "all"). The endpoint has no feature gate, so the user picks "Docling", gets a 202 + success toast, the document is created `embeddingStatus: 'pending'`, and the job sits unconsumed in Redis forever (the stuck-sweep only scans `queued`/`processing`).
   **Fix:** Add `docling-crawling` to the `WORKER_QUEUE_NAMES` value in all four templates. Add a boot-time assertion (or a unit test) that every queue in `KNOWN_WORKER_QUEUE_NAMES` appears in at least one worker-pool template, so this can't silently regress. Until the queue is provisioned everywhere, feature-gate `crawlUrlDocling` behind a flag.

3. **`DOCLING_SERVE_URL` silently defaults to `http://localhost:5001`** — `apps/api/src/external/docling-crawler/docling-crawler.constants.ts:5`
   Only the local compose stack provisions docling-serve; smoke/deploy configs point nowhere. Unlike `resolveSpiderApiKey` (which throws when unset), a misconfigured deploy just fails every `convert` call — and because those errors are swallowed (see 🟠 #5), it yields silently-empty "completed" documents.
   **Fix:** Make `resolveDoclingServeUrl` throw when unset (mirror `resolveSpiderApiKey`), or validate it in the config schema at boot. Add `DOCLING_SERVE_URL` + a `docling-serve` service to the smoke/deploy compose. Combined with #5, a missing URL then fails loudly instead of producing empty docs.

---

### 🔴 Security

4. **SSRF / local-file disclosure** — `apps/api/src/domains/documents/crawling/crawling.controller.ts:118`
   `crawlUrlDocling` validates the target only with `new URL(payload.url)` — no scheme allowlist, no private-host/IP restriction — and `CrawlUrlRequestDto` carries no server-side (class-validator) constraint, so the global `ValidationPipe` runs no checks. `new URL()` accepts `file:///etc/passwd`, `http://169.254.169.254/` (cloud metadata), `http://internal-host:6379/`. Unlike the spider engine (which forwards the URL to the external spider.cloud SaaS), the docling engine navigates it with **Chromium running inside the worker container** and stores/embeds the result. Any authenticated project member with document-create permission can read local files or reach internal services. The origin/scheme filtering in the crawler applies only to *discovered links*, never the start URL.
   **Fix:** Enforce validation server-side: make `CrawlUrlRequestDto` a class with `@IsUrl({ protocols: ['http','https'], require_protocol: true })`. In the crawler, apply the http(s)-only + private/loopback/link-local/metadata-IP block to the **start URL too**, not just discovered links, and reject redirects that land on a private host. Because DNS rebinding can defeat name-based checks, validate the *resolved IP* at connection time (custom lookup / block private CIDRs) or run the browser behind a network-isolated egress proxy.

---

### 🟠 Silent-failure & correctness

5. **Swallowed crawl errors → stored as a successful, empty document** — `docling-crawler-client.service.ts:125-128` → `docling-crawling-processor.service.ts:81`
   The per-page `catch` swallows all `goto`/`convert` failures (`errored++; continue`), so a fully-failed crawl (docling-serve down, DNS failure on the start URL) resolves with `pages=[]`. The processor has no zero-pages guard, stores `'[]'`, enqueues embeddings, and the pipeline marks the doc `completed`. The spider flow, by contrast, rejects and marks the doc `failed`. Bonus: link-enqueue runs *after* `convert` in the same `try`, so a convert failure on the start page discards its links and ends the whole crawl at zero pages.
   **Fix:** Distinguish systemic failures from per-page 404s: rethrow on connection errors to docling-serve (ECONNREFUSED) and on a failed *start* URL. If `pages.length === 0 && errored > 0`, throw so the processor's `catch` marks the doc `failed`. Add a `pages.length === 0` guard in the processor as a backstop. Move link extraction+enqueue so it isn't discarded when `convert` throws (enqueue links before attempting conversion).

6. **Cancel never cancels docling jobs; cancel-then-recrawl race** — `crawling.controller.ts:236`
   `cancelCrawl` only purges the spider queue — `BullMqDoclingCrawlingBatchService.cancelCrawlUrl` has **zero callers** (dead code). A cancelled docling crawl still runs the full (unbounded) Playwright crawl before its only guard (a post-crawl `embeddingStatus === 'failed'` check) discards it. Worse: if the user cancels then recrawls, `resetForRecrawl` sets status back to `pending`, so the stale job's guard passes, `updateContent` overwrites the fresh crawl's content, and a duplicate embeddings run is enqueued.
   **Fix:** Route `cancelCrawl` to the correct batch service by the document's persisted engine (see #7), or call both. Add a mid-crawl cancellation checkpoint in the crawler's `onPage`/loop (re-check document status or a cancel flag) so an already-running crawl stops instead of finishing. Fix the race by tagging each crawl job with a token (e.g. the BullMQ jobId) and only letting the processor write content if the document still expects *that* job — a stale job then no-ops.

7. **Recrawl always switches to spider; engine not persisted** — `crawling.controller.ts:195`
   `reCrawlUrl` always enqueues on the spider queue, and both payloads set `extractionEngine` to the same literal `'web-crawl'`, so docling docs are indistinguishable. Recrawling a docling doc silently swaps engines (docling's origin+base-path scoping → spider's whole-site `limit:0`), and in a docling-only deploy without `SPIDER_API_KEY` the worker throws *after* `resetForRecrawl` already wiped the doc, flipping it to `failed`.
   **Fix:** Persist the crawl engine on the Document (reuse the existing `extraction_engine` column or add a dedicated `crawlEngine`), set it in both `crawlUrl` and `crawlUrlDocling`, and have `reCrawlUrl` read it to enqueue on the matching queue. This single change also gives `cancelCrawl` (#6) the data it needs to route correctly.

8. **Worker stall leaves the document stuck `pending` forever** — `docling-crawling.worker.ts:12`
   `@Processor` with no options (concurrency 1, lockDuration 30s, maxStalledCount 1). A worker killed mid-crawl (deploy) stalls the job; after the stall limit BullMQ moves it to failed via the deferred path *without re-entering `process()`*, so the processor's `catch` (which would set `failed`) never runs. `@OnWorkerEvent('failed')` fires but only logs — it never touches the document (the csv-extraction/evaluation workers call `markRecordFailed`). The doc stays `pending`, the stuck-sweep skips `pending`, and the UI shows a perpetual "crawling" spinner.
   **Fix:** Add an `@OnWorkerEvent('failed')` handler that marks the document `failed` (as the csv-extraction/evaluation workers do), so deferred/stalled failures still surface. Set explicit worker options tuned for long jobs (longer `lockDuration`/`lockRenewTime`, a considered `maxStalledCount`, explicit `concurrency`). Extend the stuck-sweep to also rescue crawl docs left `pending` past a TTL.

9. **Unbounded in-process crawl** — `docling-crawler-client.service.ts:46`
   No max-page cap, no depth limit, no overall deadline (only a 30s per-goto timeout); every page's markdown accumulates in `pages[]` and the processor `JSON.stringify`s the whole array. A large same-origin site pins the concurrency-1 worker for hours and risks V8 max-string/OOM.
   **Fix:** Add a `MAX_PAGES` cap (stop enqueuing/dequeuing past it and log the truncation) and an optional max-depth and overall wall-clock deadline. For large sites, stream pages to storage incrementally rather than buffering all markdown in memory and serializing it in one shot.

10. **Browser leak on early failure** — `docling-crawler-client.service.ts:42`
    `launch()`, `newContext()`, and `newPage()` all run *before* the `try` whose `finally` is the only `browser.close()`. A rejection from `newContext()`/`newPage()` orphans the Chromium process; with `attempts=1` (no `defaultJobOptions`), leaks accumulate across user-retriggered crawls.
    **Fix:** `launch()` first, then create context/page *inside* the `try` so the `finally` always closes the browser — or wrap the whole launch+context+page acquisition in a helper that guarantees `close()` on any throw (e.g. try/finally around the full block).

11. **Duplicate crawls from fragment / start-URL dedup gaps** — `docling-crawler-client.service.ts:115`
    Dedup keys on raw hrefs with fragments intact, so `/guide#install`, `/guide#usage`, `/guide` are three entries → the same page is crawled, stored, and embedded three times. Separately, the start URL is queued as the raw `params.url` while discovered links are browser-normalized, so `https://example.com` vs `https://example.com/` re-crawls the start page.
    **Fix:** Normalize every URL before dedup: strip `parsedLink.hash` (and optionally the trailing slash), and canonicalize the start URL through `new URL()` the same way discovered links are. Key `visitedUrls` and the queue on the normalized form.

12. **`isSamePageAnchor` ignores the query string → pagination dropped** — `docling-crawler-client.service.ts:110`
    The guard is `hash !== '' && pathname === currentPathname`. A link from `/blog` to `/blog?page=2#content` matches and is treated as an on-page anchor, so all paginated content is silently missed.
    **Fix:** Once hashes are stripped for dedup (#11), this special-case largely disappears — a same-pathname/different-query URL is a distinct page and should be enqueued. If you keep the anchor check, require `search` to match too (same-page only when pathname **and** search match and a hash is present).

13. **Start-URL redirect re-scopes `basePath`, collapsing the crawl** — `docling-crawler-client.service.ts:67`
    `basePath` is re-derived from the first page's *post-redirect* URL. If `/` 301s to `/home`, `basePath` becomes `/home` and links to `/about`, `/pricing` fail `isUnderBasePath` — a whole-site crawl returns one page. Path-prefix scoping is intended and re-deriving the *origin* is necessary (http→https, apex→www), but re-deriving the *path* changes the user's requested scope; no test covers redirects.
    **Fix:** Keep `basePath` from the **requested** start URL's path and re-derive only the **origin** from the resolved URL after redirect. Add tests for path-changing redirects (`/` → `/home`, `/docs` → `/docs/`) to lock in the intended scope.

14. **Asset-extension filter tests the full href, not the pathname** — `docling-crawler-client.service.ts:117`
    The `$`-anchored regex is tested against the raw href, so `…/manual.pdf?version=2` and `…/doc.pdf#page=2` bypass it (Chromium navigates the PDF), while `…/gallery#photo.png` is wrongly skipped.
    **Fix:** Test `parsedLink.pathname` against `SKIPPED_LINK_EXTENSIONS` instead of the raw `link` string.

15. **Empty pages stored (spider skips them)** — `docling-crawler-client.service.ts:98`
    The spider client skips empty-content pages; the docling client pushes `{ url, markdown: '' }` and fires `onPage` anyway, inflating `pagesCrawled` and the stored content JSON. (The embeddings pipeline does drop empty chunks, so no wasted embedding work — but counts/content are inflated.)
    **Fix:** Mirror the spider client — skip pages whose trimmed markdown is empty before `pages.push`/`onPage`, and count them separately (e.g. `skipped`) rather than as crawled.

---

### 🟡 Reuse, efficiency & maintainability

16. **~250 lines duplicate the url-crawling vertical verbatim** — `docling-crawling-processor.service.ts` (+ worker, batch service, both modules, payload type)
    Rename-only copies (the processor differs by ~8 lines). Drift is *already* visible — the metrics service (#23) exists only on the spider side.
    **Fix:** Extract one generic processor/worker/batch module parameterized by an injected crawler-client token plus queue/job names; both `SpiderClientService` and `DoclingCrawlerClientService` already satisfy the same `crawlUrl({url,onPage})` structural interface. Alias the payload type (`export type CrawlUrlDoclingJobPayload = CrawlUrlJobPayload`). This removes ~130 lines and the duplicated spec and prevents future divergence.

17. **Frontend engine choice forked across 5 files** — `documents.thunks.ts:228` (+ spi, api, route, two middleware listeners)
    `crawlUrlDocling.fulfilled/rejected` are verbatim copies (same hardcoded `'Website crawl failed'`).
    **Fix:** Add `engine` as a parameter to the existing `crawlUrl` thunk/spi/api (the api layer picks the route, or send `engine` in the request body). Replace the duplicated listeners with `isAnyOf(crawlUrl.fulfilled, …)` / `isAnyOf(crawlUrl.rejected, …)` matchers (the file already does this for the shared refresh listener), and drop the parallel chain.

18. **O(queue) `urlQueue.includes()` per discovered link** — `docling-crawler-client.service.ts:116`
    Quadratic on large sites with repeated nav menus.
    **Fix:** Maintain a `Set<string>` of enqueued/visited URLs for O(1) membership checks; keep the array purely as the FIFO. This also removes the redundant dequeue-side dedup.

19. **`goto` and `convert` awaited serially per page** — `docling-crawler-client.service.ts:95`
    Wall time = sum of (goto + convert) per page.
    **Fix:** Pipeline the two stages — start `convert` for page N while navigating to N+1 using a small bounded in-flight set, and await/drain at the end (or when the set is full). Fetching stays sequential, so politeness is unchanged; wall time roughly halves.

20. **`cancelCrawlUrl` scans all jobs instead of a `jobId` lookup** — `bull-mq-docling-crawling-batch.service.ts:24`
    O(queue) Redis reads per cancel. (Also dead code today — see #6.)
    **Fix:** Enqueue with a deterministic `{ jobId: documentId }` (which also dedupes double-enqueues for the same document) and cancel via `(await queue.getJob(documentId))?.remove()` — the O(1) pattern already used in csv-extraction-runs, extraction-agent-sessions, and evaluations.

21. **`dl→ul` rewrite inline in the crawl loop, untested** — `docling-crawler-client.service.ts:76`
    Docling conversion preprocessing embedded in the crawler; the spec mocks `page.evaluate` wholesale so it has zero coverage, and it flattens `dd` via `textContent`, discarding nested links/lists.
    **Fix:** Extract the DOM rewrite into a named, exported, unit-tested function (or a page-injected script module) that lives with the Docling conversion path, not the generic crawler. Preserve nested content (clone child nodes) instead of `textContent`-flattening `dd`, and add a spec with representative HTML.

22. **Dead `baseUrl`/`basePath` initializers** — `docling-crawler-client.service.ts:30`
    The pre-loop values are never read (the link filter runs only after the first iteration overwrites them).
    **Fix:** Navigate `params.url` once *before* the loop, derive `const baseUrl`/`const basePath` from the resolved `page.url()`, then run the queue loop with consts. (Combines naturally with the #13 fix.)

23. **No queue metrics for the docling queue** — `docling-crawling-workers.module.ts:31`
    The spider module registers `UrlCrawlingQueueMetricsService` (OTel gauges); the docling module has no equivalent, so a stuck docling backlog goes unnoticed.
    **Fix:** Generalize `UrlCrawlingQueueMetricsService` to take a queue name and register it for the docling queue too (or emit metrics for both queues from one service). Falls out naturally if the vertical is de-duplicated (#16).

---

### 🟢 Tests & conventions

24. **Docling batch mock omits `cancelCrawlUrl`** — `test-overrides.ts:46`
    Nothing fails today (NestJS `useValue` does no conformance check), but the moment cancel is wired to the docling service — the fix for #6 — every suite using this mock throws `cancelCrawlUrl is not a function`.
    **Fix:** Add `cancelCrawlUrl: jest.fn()` to `createDoclingCrawlingBatchServiceMock` (and the spider mock, which has the same gap). Better: type the mock as the batch-service interface so a missing method fails to compile.

25. **New batch service ships without a `*.service.spec.ts`** — `bull-mq-docling-crawling-batch.service.ts:11`
    `apps/api/CLAUDE.md` mandates one. Its spider twin also lacks one (precedent), but the PR added specs for the two other new services, so this is the gap.
    **Fix:** Add `bull-mq-docling-crawling-batch.service.spec.ts` covering `enqueueCrawlUrl` and the `cancelCrawlUrl` job-matching loop (which is otherwise untested). If the vertical is de-duplicated (#16), one spec covers both engines.

26. **New engine field bypasses the mandated `Form` components** — `CrawlUrlButton.tsx:108`
    Uses `Field`/`FieldLabel` + raw `Controller` + inline `z.object` instead of the shared `Form`/`FormField`/`FormMessage` + api-contracts schema. Low practical impact (the field is a `z.enum` with a valid default, so it can't produce a visible error), but it extends the form's divergence from ADR 0012.
    **Fix:** Migrate the field — ideally the whole form — to `Form`/`FormField`/`FormControl`/`FormMessage` with a `zodResolver` schema exported from `@caseai-connect/api-contracts`. Low urgency; best done as a small follow-up that also removes the hand-written error `<p>` for `url`.

27. **Unrelated drive-by hunks** — `apps/web/src/stories/routes/studio/AgentEditorRoute.stories.tsx:139` and `apps/help/.../VersionHistoryWalkthrough.astro`
    A whitespace-only edit and a comment-only rename belonging to the #629 agentSettings refactor.
    **Fix:** Revert both hunks from this PR; land the astro comment rename as its own small chore commit so blame/revert of the crawler feature stay clean.

---

**Suggested merge order:** the three deploy blockers (#1–#3) and the SSRF gap (#4) are must-fix; the silent-failure trio (#5, #6, #8) turns real failures into invisible ones and is close behind. Persisting the chosen engine on `Document` (#7, via the existing `extraction_engine` column) fixes recrawl *and* cancel routing (#6) at once, and collapsing the duplicated vertical (#16/#17) would have prevented most of the divergence — and folds in the metrics (#23) and spec (#25) gaps for free.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

apps/api/src/common/utils/url-safety.ts
did, line 1: the `common/utils` folder shouldn't reference any functionality. 
Having `crawl-url` in the name of the file doesn't following this guide line.

That being said, the 2 functions can be used for any kind of purposes, not specifically CrawUrls. So, as a solution, I'd remove the reference of `crawl` from the file name and the function names.

PR #659, opened 2026-08-24
fix(api): split CORS policies between public embed and platform endpoints

### LINKED ISSUE #366: Public API and cors
- useless `const _frontendUrls = parseFrontendUrls(process.env.FRONTEND_URL)`
- `process.env.FRONTEND_URL` should be removed
- ```app.enableCors({
    origin: (origin, callback) => callback(null, origin ?? true)``` It is ok ? @jdoucy 

Perhaps we need a public API ?? @jdoucy

jdoucy:
Closes #366

## Summary

The API reflected any request origin on every endpoint with `credentials: true`. This splits CORS into two policies, selected per request by a delegate:

- `/public/*` (embed widget) keeps reflecting the origin — security is enforced by `EmbedTokenGuard` (embed token + per-config allowedOrigins), not CORS.
- Everything else pins origins to `FRONTEND_URL`. This also covers the one cookie-authenticated surface (Bull Board's OIDC session).
- Credentialed CORS is dropped: no caller sends cookies or uses `credentials: 'include'` (verified across apps/web and the embed).

The previously dead `parseFrontendUrls` call in `main.ts` (first point of the issue) becomes the real origin list. The logic moves to `src/config/cors.ts` with an 11-test spec (parsing, public/strict split, the `/publications` look-alike trap, no credentials). In production, an unset or blank `FRONTEND_URL` throws at bootstrap: the Cloud Run revision fails and the previous one keeps serving, instead of a healthy API that silently blocks every platform call.

## ⚠️ Before merging

The deploy is dispatched to the infra repo — verify `FRONTEND_URL` is set on the Cloud Run service there. If it is unset, the API refuses to start.

## Test plan

- `npx jest src/config/cors.spec.ts` — 11 tests green
- `make tests-parallel` — full suite green
- typecheck + biome green

### DIFF
--- a/.claude/skills/end-feature/SKILL.md
+++ b/.claude/skills/end-feature/SKILL.md
@@ -8,12 +8,14 @@ Wrap up a feature by running its final steps. Present a checklist form, then exe
 
 ## Step 0 — Show the checklist form
 
-Before doing anything, call `AskUserQuestion` with a single **multi-select** question so the user can check/uncheck the steps. All four steps are pre-selected by default (list them as the options; the user unchecks any they want to skip):
+Before doing anything, call `AskUserQuestion` with a single **multi-select** question so the user can check/uncheck the steps (list them as the options; the user unchecks any they want to skip):
 
 - **Update changelog** — add an entry under `## [Unreleased]` in `CHANGELOG.md`
 - **Create a branch** — move current work onto a feature branch
 - **Run tests-parallel** — `make tests-parallel`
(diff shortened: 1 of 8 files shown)


did:
@jdoucy This looks good to me. I'm the one to blame, I removed FRONTEND_URLS from the Cors ([#294](https://github.com/bayesimpact/bayes-platform/pull/294)). Sorry :-(

PR #666, opened 2026-08-25
feat(llm): send pdfs as page images to Gemma and MedGemma models

### LINKED ISSUE #322: Allow pdf for gemma models
Gemma models needs images, we have to transform pdfs into images

alexisjamet:
## Summary

Gemma-family vLLM models are image-only, so a PDF attachment previously threw `Model cannot process application/pdf file` (extraction) or failed at vLLM (chat). This PR rasterizes each PDF page to a PNG image part before the LLM call, for the Gemma and MedGemma providers, in both `generateStructuredOutput` and `streamChatResponse`. Mistral keeps its existing PDF rejection.

**Rasterization runs in a dedicated service** (`apps/pdf-renderer`), not in the API/workers: rendering is RAM-heavy (up to ~1GB per document), so 100 concurrent conversions must not compete with the main API for memory. The API calls the service over HTTP and keeps the exact same `convertPdfPartsToImageParts` contract; `pdfjs-dist` and `@napi-rs/canvas` are removed from `apps/api` entirely. Deployment (Cloud Run, 2Gi instances, per-instance concurrency 2, scale-out) is in bayesimpact/infra#115.

Closes #322

## The dedicated pdf-renderer service

- **`POST /render-pages`** — raw `application/pdf` body + optional `maxPages` / `maxPixelsPerPage` / `scale` query params (the API sends its own limits; the service enforces safe defaults and hard ceilings) → `{"pages": ["<base64 png>", ...]}`. **`GET /healthz`** for probes.
- **Auth**: Cloud Run invoker IAM — in production `roles/run.invoker` is restricted to the platform's api service account, and the API client mints a Google ID token for the service URL when `PDF_RENDERER_AUTH=google-iam` (set by Terraform). Google rejects unauthenticated requests before they reach the container; no shared secret to store or rotate. Locally the renderer runs open and no header is sent.
- **Cloud Run caps handled explicitly**: requests are pre-checked client-side against the 32MiB HTTP/1 request limit (clear error instead of an opaque frontend 413), and responses are **streamed chunked** because a 20-page scanned PDF measured 78MiB of base64 JSON — over the 32MiB non-chunked response cap.
- **API client** (`apps/api/src/external/llm/pdf-to-image-parts.ts`): `PDF_RENDERER_URL` / `PDF_RENDERER_APIKEY` env vars, 60s timeout, propagates the service's error messages (page-limit message unchanged). Fails with an explicit configuration error when the URL is unset.
- **Local dev**: `npx turbo dev --filter=@caseai-connect/pdf-renderer` (port 3001) + `PDF_RENDERER_URL=http://localhost:3001` in `apps/api/.env` (documented in `.env-example` and the app README).
- **CI**: renderer tests wired into `make tests-only-parallel`; own Dockerfile built by `make docker-build` and trivy-gated in the Security workflow; `apps/pdf-renderer` paths added to the deploy-trigger `API_PATHS`.

## Package choices

(now scoped to `apps/pdf-renderer` — these deps left `apps/api`)

- **`pdfjs-dist@6.2.108` (pinned exactly)** — Mozilla pdf.js, the PDF engine shipped in Firefox: heavily fuzzed, fast patch cadence, only 3 advisories ever (all JS-injection class, all fixed; 6.2.108 clears all of them, including CVE-2026-16633 published 2026-08-06). Pinning the engine directly keeps its version visible to `npm audit`/Dependabot/Trivy.
- **`@napi-rs/canvas@1.0.8`** — the canvas backend pdf.js itself declares as its only `optionalDependency` (deduped in our tree). Best-in-class supply-chain hygiene: no install scripts anywhere in the family, prebuilt statically-linked Skia (no system libs, verified with `ldd`), npm provenance/SLSA attestations from GitHub-hosted CI, 21M weekly downloads, maintained by the NAPI-RS author.
- **Why not `unpdf`** (the first iteration): it re-bundles pdf.js invisibly to scanners, and its current release still bundles pdf.js 6.1.200, which is inside the CVE-2026-16633 affected range — the 2024 pdf.js RCE fix also took ~6 weeks to reach an unpdf release. Note `unpdf@0.12.2` remains in the tree transitively via the deprecated `@llamaindex/readers` (pre-existing text-extraction path, its bundled pdf.js 4.6.82 is advisory-clean); migrating off `@llamaindex/readers` is a separate follow-up.

## Tech choices

- **Rendering still runs in a short-lived subprocess** inside the service (`pdf-pages-to-png.script.mjs`, spawned per request): a malicious or degenerate PDF can only crash/hang a killable child, never the serving event loop. The child gets a 60s hard timeout, a 1GB V8 heap cap, and an **empty environment** (no secrets reachable from the renderer). A child that dies before draining stdin surfaces as an HTTP error instead of an EPIPE process crash (found by adversarial review, reproduced, fixed, covered). Exit code 2 = page-limit exceeded → HTTP 422 with the same user-facing message as before.
- **Resource limits**: max 20 pages per PDF (explicit error above), per-page pixel budget of 4MP (a PDF can declare an arbitrarily large MediaBox; rendering allocates width×height×4 bytes, so the scale is clamped instead of trusting the document), render scale 2 (≈144dpi) for legible text. Per-instance concurrency is capped at 2 in Cloud Run so instances scale out instead of stacking renders.
- **nest-cli `assets` rule** ships `*.script.mjs` to `dist` so the same relative path works in dev (src), tests, and production (dist) — now in `apps/pdf-renderer` (reverted in `apps/api`).

## Alternatives considered and rejected

- **poppler `pdftoppm` subprocess**: viable (distro-patched, isolated) but needs a Dockerfile change plus poppler on every dev machine, and adds a C++ parser surface with a steady CVE stream.
- **`@hyzyla/pdfium` (WASM) + pngjs**: real memory-safety sandbox, but trusts a solo maintainer's PDFium repackaging.
- **mupdf npm**: AGPL, disqualifying for a proprietary SaaS.
- **Docling / GPU worker offload**: Docling parses PDFs with native PDFium (`pypdfium2`) and IBM's C++ parser — it relocates the attack surface rather than reducing it, and Docling is disabled in the API/CPU-worker images where the conversion previously ran inline.
- **sharp**: prebuilt libvips deliberately excludes PDF support.
- **In-process rendering in the API** (this PR's first iteration): worked, but let PDF spikes contend with API/worker memory — hence the dedicated service.

## Testing

- **Service** (10 e2e tests, `apps/pdf-renderer/src/render/render.spec.ts`, real HTTP + real subprocess rendering): single/multi-page conversion, page cap (default + per-request), pixel-budget clamp, invalid pdf/body/params, chunked transfer encoding. (No app-level auth: production access control is Cloud Run invoker IAM.)
- **API client** (14 tests, `pdf-to-image-parts.spec.ts`, against a stub HTTP server): part replacement/pass-through, exact bytes/query/auth-header forwarding, URL/base64/buffer inputs, error propagation (page-limit message, non-JSON errors, unreachable service, missing config, 32MiB pre-check), Google ID-token header in `google-iam` mode.
- **Live locally**: service rendered the real PDF fixture via curl (201, chunked, valid 1190×1684 PNG) and the actual API client code exercised against the running service.
- Live end-to-end against the real Gemma 4 26B vLLM (pre-extraction iteration): correct values extracted from the PDF fixture. The env-gated `ai-sdk-gemma.provider.spec.ts` PDF case matches the implementation.
- Adversarial multi-agent review over both repos: 4 confirmed findings, all fixed (stdin EPIPE crash, missing trivy gate for the new image, both Cloud Run 32MiB caps).
- Full gates green: `biome:ci`, repo-wide `typecheck`, `check:boundaries`, `test:parallel` (full suite, exit 0), `terraform validate` on both infra envs.

## Deploy notes

- Companion infra PR: bayesimpact/infra#115 (service + env wiring + IAM invoker binding; nothing to create out of band — no secrets). The renderer is **not** part of the automatic deploy pipeline (it changes rarely): deploy it manually via the infra repo's **"Deploy PDF Renderer"** GitHub action (or `make deploy-pdf-renderer REGION=eu PROJECT=<connect|health> version=<sha>` locally) — run it once per project before Gemma/MedGemma PDF requests hit production. `apps/pdf-renderer` changes no longer trigger the platform deploy dispatch.


apps/api/src/external/llm/ai-sdk-llm-provider-base.ts
```diff
@@ -740,11 +748,12 @@ ${leakedCall.raw}`,
...
-    //Gemma and Mistral restriction: no pdf
-    if (
-      AgentModelToAgentProvider[config.model] === AgentProvider.Gemma ||
-      AgentModelToAgentProvider[config.model] === AgentProvider.Mistral
-    ) {
+    // Gemma and MedGemma only accept images: send pdfs as one image per page
+    if (modelRequiresPdfAsImages(config.model)) {
+      message = await convertPdfPartsToImageParts(message)
+    }
+    //Mistral restriction: no pdf
```
alexisjamet, line 755: @jdoucy Dois-je utiliser convertPdfPartsToImageParts aussi pour Mistral ou on n'autorise pas les PDF du tout sur Mistral?

PR #675, opened 2026-08-27
feat(llm)!: gcs-native pdf-converter service and stable page image urls for Gemma/MedGemma

alexisjamet:
Follow-up to #666 / related to #322. Replaces the in-API PDF→image byte pump that OOM-killed the production API with a GCS-native pipeline: bytes flow GCS → converter → GCS, and image-only models fetch page images themselves via V4-signed GCS URLs.

Spec: `docs/superpowers/plans/2026-08-26-pdf-page-images-spec.md` · Plan: `docs/superpowers/plans/2026-08-26-pdf-page-images-pipeline.md` — note the spec still describes the earlier **302-redirect** design; the implementation moved to **direct signed GCS URLs** in fea14e1b.

## Problem

For Gemma/MedGemma the API downloaded whole PDFs from GCS into heap, POSTed them to `apps/pdf-renderer`, parsed a giant base64-JSON of pages, and re-encoded everything as `data:` URLs — hundreds of MB of transient allocations per conversion, no caching, Cloud Run OOM (503s incl. CORS preflights).

## How it works now

1. **New Go service `apps/pdf-converter`** (go-pdfium, WebAssembly/wazero backend — `CGO_ENABLED=0`, malicious PDFs sandboxed). `POST /render-document` downloads the PDF from GCS, enforces the page limit before rendering, renders each page (scale 2, ≤4M px/page), uploads `{org}/{proj}/derived/{id}/page-{n}.png`, returns `{"pageCount": n}`. `POST /page-count` returns the page count without rendering. Errors map to 400/404/413/422 with `{"message"}`.
2. **Page-limit fail-fast**: the API pre-flights `/page-count`; a PDF over 20 pages throws `PdfPageLimitExceededError` before any rendering. Extraction runs record `errorCode: PDF_PAGE_LIMIT_EXCEEDED`; chat surfaces the message verbatim in the error bubble.
3. **Eager render + Postgres cache**: new nullable `pdf_page_count` column on `agent_message_attachment_document` and `document`; the converter is called once per document at LLM-request build time, then skipped forever.
4. **Signed-URL passthrough** (replaces the redirect endpoints): message building emits `{type: "image", image: URL}` parts pointing at 15-minute V4-signed GCS URLs; vLLM fetches the pages server-side. MedGemma's custom model got `supportedUrls` (restricted to `storage.googleapis.com` as SSRF defense-in-depth) + an `image_url` branch; Gemma's stock OpenAI-compat model passes URLs through natively.
5. **Byte pump and `apps/pdf-renderer` deleted in this PR** (breaking): `convertPdfPartsToImageParts`, the pdf-renderer fetch path and the 32MiB cap are gone; `getTemporaryUrl` now signs V4.

## Env vars (new)

`PDF_CONVERTER_URL`, `PDF_CONVERTER_AUTH` (`google-iam` in prod) on the API/workers; `GCS_STORAGE_BUCKET_NAME`, `PDF_CONVERTER_MAX_PDF_BYTES`, `PORT` on the converter. `API_PUBLIC_BASE_URL` is **no longer needed** — it belonged to the dropped redirect design. See `apps/api/.env-example` and `apps/pdf-converter/README.md`.

## Testing locally

The converter is GCS-native, so the Gemma/MedGemma PDF path needs a real bucket: run `go run .` in `apps/pdf-converter` with `GCS_STORAGE_BUCKET_NAME` + ADC credentials, and set `PDF_CONVERTER_URL=http://localhost:3002` **and `GCS_STORAGE_BUCKET_NAME` on the API too** — with LocalStorageService (no API bucket) the PDF→image path fails with the converter's 404, since the source PDF only exists on local disk (see review findings). All other flows are unchanged.

## Verification status

- `biome:check`, `typecheck`, `check:boundaries`, `go vet` + `go test` (converter unit + server tests): all green locally.
- `make tests-parallel` could not complete locally: the shared `connect_test` DB has pre-existing schema drift (columns from two **main** migrations present but unrecorded in the ledger), so `migration:test:run` fails before any test runs — on main too. Fix is recreating `connect_test`; CI runs on a fresh DB and is unaffected.

## Review findings

An automated deep review (11 finder angles, adversarially verified) posted **31 findings as PR comments, each tagged with severity: 2 high / 7 medium / 22 low**. The two high ones:

- **Converter has no render/request deadline** — a PDF that hangs pdfium permanently occupies one of the 4 pool instances while `/healthz` stays green; four such renders wedge the service until manual restart.
- **MedGemma chat streaming drops image parts** — `formatMessages` in the custom model flattens content to text (pre-existing on main), so the chat-attachment flow cannot work for MedGemma agents with tools until that path is fixed.

## Post-merge / infra checklist (infra repo)

- [ ] Terraform: `pdf-converter` Cloud Run service (1–2GB RAM, min 0), SA with `roles/storage.objectAdmin` on the bucket, `roles/run.invoker` for API/workers SA; API/workers env: `PDF_CONVERTER_URL`, `PDF_CONVERTER_AUTH=google-iam`.
- [ ] "Deploy PDF Converter" GitHub action (clone of Deploy PDF Renderer).
- [ ] Smoke-test prod egress: Gemma/MedGemma chat completion with `image_url` → a V4-signed `storage.googleapis.com` URL.
- [x] ~~Decommission `apps/pdf-renderer` in its own PR once the converter is live~~ — removed in this PR (bb60af3a).


docs/superpowers/plans/2026-08-26-pdf-page-images-pipeline.md
```diff
@@ -0,0 +1,1690 @@
...
+            // Image-only models: send one rendered page image URL per page. The
+            // pages live in GCS (rendered once by pdf-converter, cached via
+            // pdf_page_count) and the model fetches them through the public
+            // pdf-pages redirect endpoint — no pdf/image bytes in this process.
+            if (!hasStorageBucketName) {
+              throw new Error(
+                "PDF attachments with Gemma/MedGemma models require GCS storage and the pdf-converter service (set GCS_STORAGE_BUCKET_NAME and PDF_CONVERTER_URL)",
+              )
+            }
+            const pdfPageCount = await this.pdfPagesService.ensureRenderedPages({
```
alexisjamet: - call pdf converter client to know number of pages (without creating images)
- if page count over limit do nothing and return frontend error
- if page count inside limit then create images

apps/api/src/domains/documents/document-pdf-pages.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Public, unauthenticated capability endpoint (same pattern as
+ * ResourceLibraryFilesController): keyed by the document UUID and scoped to
+ * the org/project in the path, it only ever signs derived page objects under
+ * that project's own prefix. 302-redirects to a freshly signed GCS URL so the
+ * LLM serving stack (vLLM fetches image_url server-side, follows redirects,
+ * cannot send auth headers) can load pdf page images.
+ */
+@Controller()
+export class DocumentPdfPagesController {
```
alexisjamet: fix: remove this. give signed GCS url directly to LLM

apps/api/src/domains/agents/shared/agent-session-messages/streaming/agent-llm-request.service.ts
```diff
@@ -276,12 +280,52 @@ export class AgentLlmRequestService extends ServiceWithLLM {
...
+                attachmentDocumentId,
+                connectScope,
+                pdfPageCount,
+              })
+            }
+            const content = llmMessage.content as Array<ImagePart>
+            for (let pageNumber = 1; pageNumber <= pdfPageCount; pageNumber++) {
+              content.push({
+                type: "image",
+                image: this.pdfPagesService.buildAttachmentPageImageUrl({
```
alexisjamet: fix: give signed GCS url directly to LLM

alexisjamet (commented):
Automated deep review of this PR: 11 finder angles → dedup → adversarial verification pass; every finding below was confirmed against the actual code (including go-pdfium, TypeORM, GCS client, and google-auth-library sources).

**31 findings: 2 🔴 high · 7 🟠 medium · 22 🟡 low.** 29 are inline comments; 2 anchor to code outside the diff (derived-page retention, presigned-upload overwrite) and follow as regular PR comments. Each comment carries its severity status.

The PR description has been updated to match the implemented design (direct signed GCS URLs; the 302-redirect design and `API_PUBLIC_BASE_URL` are gone) and now lists the two high findings under “Review findings”.

apps/pdf-converter/internal/render/render.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+	if err != nil {
+		return 0, fmt.Errorf("%w: %v", ErrInvalidPdf, err)
+	}
+	return pageCountResponse.PageCount, nil
+}
+
+// RenderPages rasterizes every page as PNG and hands each to emit with a
+// 1-based page number. Returns the page count.
+func (renderer *Renderer) RenderPages(
```
alexisjamet, line 121: **Severity: 🔴 High** · confirmed by adversarial verification

**No render deadline: a hung pdfium render permanently wedges the pool.** `RenderPages`/`GetPageCount` take no context, wazero's close-on-context-done is off, and the deferred `instance.Close()` never runs if the call never returns. With `MaxTotal: 4`, four hung renders (or BullMQ retries of one hostile PDF) make every subsequent request fail `get pdfium instance` after the 30s pool-acquire timeout — while `GET /healthz` never touches the pool and stays green, so Cloud Run never restarts the instance. The API client's comment that "the converter's own render timeout is stricter" is currently false: there is no render timeout at all.

**Fix:** run the render under a `context.WithTimeout` (e.g. 60s like the old pdf-renderer's hard kill) with wazero configured to interrupt on context cancellation, and/or make `/healthz` acquire-and-release a pool instance so a wedged pool fails health checks.

apps/api/src/external/llm/providers/medgemma/custom-med-gemma-language-model.ts
```diff
@@ -198,6 +197,13 @@ export class CustomMedGemmaLanguageModel implements LanguageModelV3 {
           case "file": {
             if (!value.mediaType.startsWith("image/"))
               throw new Error(`MedGemma model cannot process ${value.mediaType} file`)
+            if (value.data instanceof URL) {
```
alexisjamet, line 200: **Severity: 🔴 High** · confirmed by adversarial verification

**MedGemma chat streaming silently drops these image parts.** This URL branch only helps `doGenerate` (structured output). The streaming path — `doStreamWithTools`, which every real conversation agent hits since agents always get a ToolSet, and the only streaming variant this model implements — goes through `formatMessages` (line ~268), which flattens user content with `msg.content.map((c) => c.text || "").join("")`, reducing every `image_url` part to an empty string. A user who attaches a PDF to a MedGemma chat agent gets confident answers about a document the model never received, with no error anywhere. The flattening pre-exists on main (it dropped the old inline page bytes too), but this PR's chat-attachment flow ships through a path where it demonstrably cannot work, and no test asserts what the wrapper actually transmits (the new spec covers `doGenerate` only).

**Fix:** make `formatMessages` preserve image parts (mirror the `toInput` URL/image handling) and add a spec asserting the transmitted request body for `doStreamWithTools`.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
```diff
@@ -0,0 +1,122 @@
...
+          `pdf-converter request to /${path} timed out after ${RENDER_REQUEST_TIMEOUT_MS}ms`,
+        )
+      }
+      const reason = error instanceof Error ? error.message : String(error)
+      throw new Error(`could not reach pdf-converter at ${converterUrl}: ${reason}`)
+    }
+    if (!response.ok) {
+      throw new Error(await this.extractErrorMessage(response))
+    }
+    return (await response.json()) as { pageCount: number }
```
alexisjamet, line 87: **Severity: 🟠 Medium** · confirmed by adversarial verification

**Unvalidated response shape propagates `undefined` silently.** `(await response.json()) as { pageCount: number }` — if a misconfigured `PDF_CONVERTER_URL`/proxy returns 200 with a different JSON shape, `pageCount` is `undefined`, which passes the limit check (`undefined > 20` is false), then `onPageCountUpdate(undefined)` **silently no-ops** in both persistence paths (verified on TypeORM 0.3.30: undefined values are skipped and the `@UpdateDateColumn` turns it into a touch-update — no throw), and `Array.from({length: undefined})` yields `[]` — the LLM request goes out with zero page images and no error. Also: a 200 with a non-JSON body surfaces a bare `SyntaxError`, and this line sits outside the try/catch, so a timeout firing during body read escapes as a raw `TimeoutError` instead of the descriptive message.

**Fix:** validate the parsed body (`typeof pageCount === 'number' && Number.isInteger(pageCount) && pageCount >= 0`) and move the parse inside the mapped try/catch.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    document: { storageRelativePath, pdfPageCount },
+    onPageCountUpdate,
+    fileStorageService,
+  }: {
+    document: { storageRelativePath: string; pdfPageCount: number | null }
+    onPageCountUpdate: (pdfPageCount: number) => Promise<void>
+    fileStorageService: IFileStorage
+  }): Promise<string[]> {
+    if (pdfPageCount === null) {
+      // If the PDF page count is not known, render the document to determine it.
```
alexisjamet, line 38: **Severity: 🟠 Medium** · confirmed by adversarial verification

**`pdf_page_count` is a write-once cache with no invalidation path.** No code ever resets it to null (`resetForRecrawl` doesn't touch it), and V4 signing is pure crypto that never checks the object exists. If the `derived/` page objects are ever removed (bucket lifecycle rule, cleanup, migration, layout change), every future Gemma/MedGemma run on that document signs dead URLs that 404 at vLLM's server-side fetch — an opaque provider error, forever, recoverable only by manually nulling the DB column.

**Fix:** on a failed model fetch (or on a converter 404), null the cache and re-render once; or have the converter check the output prefix idempotently instead of trusting the DB count.

apps/api/src/domains/documents/storage/gcs-storage.service.ts
```diff
@@ -52,7 +52,9 @@ export class GcsStorageService implements IFileStorage {
     const bucket = this.storage.bucket(this.bucketName)
     const file = bucket.file(storageRelativePath)
     const expires = Date.now() + 15 * 60 * 1000 // 15 minutes
```
alexisjamet, line 54: **Severity: 🟠 Medium** · confirmed by adversarial verification

**15-minute expiry is shorter than the LLM calls that consume these URLs.** Extraction runs explicitly opt into 30-minute network timeouts (`useExtendedTimeouts` → the vertex dispatcher), and chat tool loops re-send the same image URLs on every step (up to `stepCountIs(20)`), with vLLM re-fetching them server-side each round-trip. A queued/cold endpoint or a tool loop past the 15-minute mark gets a GCS 403 mid-turn, surfacing as an opaque provider error; a BullMQ retry can consume the same window again. The sibling `generateSignedUploadUrl` already takes `expiresInSeconds` as a parameter, so the fix shape exists ten lines down.

**Fix:** add an expiry parameter to `getTemporaryUrl` and sign page-image URLs for ≥ the extended request timeout (e.g. 45–60 min), keeping 15 min for user download links.

apps/api/src/domains/documents/pdf-pages/pdf-page-limit-exceeded.error.ts
```diff
@@ -0,0 +1,12 @@
+// Thrown before rendering when a pdf has more pages than image-only models
+// accept. The message is user-facing: it is shown verbatim in the chat error
```
alexisjamet, line 2: **Severity: 🟠 Medium** · confirmed by adversarial verification

**This comment overstates the extraction UX: the message never reaches the extraction UI.** The runner stores `errorCode`/`errorDetails` and the DTO carries them into the web model, but no component renders them for extraction sessions — a failed run shows only a 'failed' badge and a generic 'Extraction failed' toast (the `t("error")` fallback in `ExtractionAgentSessionItem` is only reachable for *successful* runs). Only the **chat** path shows the message verbatim (SSE error event → error bubble). So the PR's stated fail-fast-with-user-facing-error purpose (952433c8) is delivered for chat only; an extraction user retries the same oversized PDF with no explanation. The rendering gap itself pre-exists (same for `SCHEMA_VALIDATION_FAILED`), but this PR adds the code and the comment.

**Fix:** render `errorDetails` (or a translated `errorCode` message) on failed extraction session items, or correct this comment.

apps/pdf-converter/internal/render/render.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return 0, fmt.Errorf("get size of page %d: %w", pageIndex+1, err)
+		}
+		scale := renderScale
+		if size.Width*size.Height*scale*scale > float64(maxPixelsPerPage) {
+			scale = math.Sqrt(float64(maxPixelsPerPage) / (size.Width * size.Height))
+		}
+		rendered, err := instance.RenderPageInPixels(&requests.RenderPageInPixels{
+			Page:   page,
+			Width:  int(size.Width * scale),
```
alexisjamet: **Severity: 🟠 Medium** · confirmed by adversarial verification

**Truncation to 0 here fails the whole document — or silently bypasses the pixel budget.** `int(size.Width*scale)`/`int(size.Height*scale)` have no floor: a degenerate page under 0.5pt in both dimensions yields 0×0 and go-pdfium errors (`no width or height given`), turning one bad page into a 500 for the entire document. Worse nuance (verified in go-pdfium v1.19.8): if the pixel-budget clamp zeroes exactly **one** dimension, go-pdfium treats 0 as "not set" and recomputes it from the page aspect ratio — silently bypassing `maxPixelsPerPage` (e.g. a 1e8×1pt page renders ~20M px against the 4M budget).

**Fix:** clamp both dimensions to `max(1, ...)` after scaling.

apps/pdf-converter/store.go
```diff
@@ -0,0 +1,64 @@
...
+func (store *gcsStore) Download(ctx context.Context, object string) ([]byte, error) {
+	reader, err := store.bucket.Object(object).NewReader(ctx)
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		return nil, errObjectNotFound
+	}
+	if err != nil {
+		return nil, fmt.Errorf("open %s: %w", object, err)
+	}
+	defer reader.Close()
+	data, err := io.ReadAll(reader)
```
alexisjamet, line 46: **Severity: 🟠 Medium** · confirmed by adversarial verification

**Unbounded `io.ReadAll` makes the size cap advisory (stat-then-download race).** `Size()` uses a separate `Attrs` call and `Download()` re-opens the object with no generation pinning, so an object replaced with a larger one between the two calls is fully buffered regardless of `maxSourceBytes` — a multi-GB blob would OOM the instance mid-request. `server_test.go` covers the deleted-between-stat-and-download case but not replaced-with-larger. Mitigated by the bucket being writable only by trusted services, hence medium.

**Fix:** wrap the reader in `io.LimitReader(r, maxSourceBytes+1)` (erroring when exceeded), or pin `Generation` from the Attrs call — either makes the cap authoritative and also removes the double round trip.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
```diff
@@ -0,0 +1,122 @@
...
+  // (user-facing message) without rendering anything when the document has
+  // more pages than image-only models accept.
+  async generatePdfPageImages({
+    sourceObject,
+    outputPrefix,
+  }: {
+    sourceObject: string
+    outputPrefix: string
+  }): Promise<number> {
+    const pageCount = await this.getPageCount({ sourceObject })
```
alexisjamet, line 45: **Severity: 🟡 Low** · confirmed by adversarial verification

**The `/page-count` pre-flight doubles the converter's work, and the render-time 422 is misclassified.** Each Go endpoint independently stats, fully downloads (≤50MB), and pdfium-parses the same PDF, so every first render costs 2 downloads + 2 parses + 2 ID-token mints — while `/render-document` already fails fast on the limit *before* rendering. And the typed `PdfPageLimitExceededError` only exists via this client-side comparison: if the render-time 422 fires (pre-flight bypassed or object changed between calls), it surfaces as a generic `Error`, so the extraction runner records `EXTRACTION_PROVIDER_ERROR` instead of `PDF_PAGE_LIMIT_EXCEEDED`. Drift between the two checks is impossible (the server takes `maxPages` from the request body), and `/page-count` has no other caller.

**Fix:** drop the pre-flight and map the render 422 to `PdfPageLimitExceededError` (the server could include `pageCount` in the 422 body), deleting the endpoint and client method.

apps/api/src/domains/agents/shared/agent-session-messages/streaming/agent-llm-request.service.ts
```diff
@@ -263,44 +270,56 @@ export class AgentLlmRequestService extends ServiceWithLLM {
...
-          )
-
-          const content = llmMessage.content as Array<FilePart>
-          content.push({
-            type: "file",
-            mediaType: "application/pdf",
-            data,
-            filename: attachmentDocument.fileName,
-          })
+          if (modelRequiresPdfAsImages(model)) {
```
alexisjamet, line 281: **Severity: 🟡 Low** · confirmed by adversarial verification

**Local dev without a GCS bucket now always fails on this path.** With `GCS_STORAGE_BUCKET_NAME` unset (the `.env-example` default) the API stores files via `LocalStorageService`, but the converter is GCS-only, so `/page-count` 404s `source pdf not found`; the old `hasStorageBucketName` public-sample-URL fallback was deleted with no replacement, and nothing warns at startup or call time about the LocalStorage+GCS-converter mismatch. Dev-only impact (the PR description's local-testing section now documents the requirement).

**Fix:** a startup warning (or clear error) when `PDF_CONVERTER_URL` is set while the API uses LocalStorageService.

apps/api/src/domains/documents/documents.service.ts
```diff
@@ -208,6 +208,30 @@ export class DocumentsService {
...
+  async updatePdfPageCount({
+    connectScope,
+    documentId,
+    pdfPageCount,
+  }: {
+    connectScope: RequiredConnectScope
+    documentId: string
+    pdfPageCount: number
+  }): Promise<void> {
+    await this.documentRepository.update(
```
alexisjamet, line 225: **Severity: 🟡 Low** · confirmed by adversarial verification

**Hand-rolled connect scoping + missing `affected` check.** `apps/api/CLAUDE.md` says services must delegate scoping to `ConnectRepository` rather than re-implement it in where clauses — `documentConnectRepository.updateOneById` is available on this service, and the sibling `AgentMessageAttachmentDocumentsService.updatePdfPageCount` added in this same PR uses exactly that. Also, unlike every neighbor (`updateEmbeddingStatus`, `markAsUploaded`, `updateContent`, `resetForRecrawl`), this never checks `result.affected`, so a scope mismatch or soft-deleted row (the raw criteria don't exclude `deleted_at`) is a silent no-op — the count is never cached and the converter re-renders every run with no surfaced error. Practically unreachable today (the only caller passes a scope-checked relation), hence low.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  async getImageUrls({
+    document: { storageRelativePath, pdfPageCount },
+    onPageCountUpdate,
+    fileStorageService,
+  }: {
+    document: { storageRelativePath: string; pdfPageCount: number | null }
+    onPageCountUpdate: (pdfPageCount: number) => Promise<void>
+    fileStorageService: IFileStorage
+  }): Promise<string[]> {
+    if (pdfPageCount === null) {
```
alexisjamet, line 37: **Severity: 🟡 Low** · confirmed by adversarial verification

**No single-flight around render-then-cache: concurrent first renders can fail spuriously.** Two runs on a fresh document both see `pdfPageCount === null` and both render to the same object names. The converter's uploads set no preconditions, and precondition-free GCS writes are non-idempotent and **not retried** by the Go storage client (verified in cloud.google.com/go/storage v1.65.0) — a same-object-write 429 (GCS caps ~1 mutation/s/object) becomes writer error → HTTP 500 → one run fails with `EXTRACTION_PROVIDER_ERROR` even though all pages exist and the sibling run succeeded. Self-correcting on retry, plus doubled render cost.

**Fix:** a per-document advisory lock or in-flight de-dup around the render, or converter-side idempotency (skip upload if the prefix is already complete).

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+ * pdf_page_count column. The model fetches pages through temporary signed
+ * storage URLs, so no image bytes ever transit this process.
+ */
+@Injectable()
+export class PdfPagesService {
+  constructor(private readonly pdfConverterClient: PdfConverterClient) {}
+
+  derivedPagesPrefix(storageRelativePath: string): string {
+    const lastSlashIndex = storageRelativePath.lastIndexOf("/")
+    const directory = storageRelativePath.slice(0, lastSlashIndex)
```
alexisjamet, line 19: **Severity: 🟡 Low** · confirmed by adversarial verification

**`lastIndexOf('/') === -1` chops the last character.** For a slash-less `storageRelativePath`, `slice(0, -1)` truncates the filename: `'doc1.pdf'` → directory `'doc1.pd'` → prefix `'doc1.pd/derived/doc1/'`. Latent today (all production paths are `{org}/{proj}/{id}.{ext}`), but writes and reads share this function, so a future bare-name caller (seed script, import) plants derived trees at corrupt top-level prefixes — outside any org/project folder, invisible to per-tenant cleanup — and nothing ever fails loudly because signing points at the same wrong prefix.

**Fix:** handle `-1` explicitly (empty directory) or assert the expected path shape.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+      // If the PDF page count is not known, render the document to determine it.
+      pdfPageCount = await this.pdfConverterClient.generatePdfPageImages({
+        sourceObject: storageRelativePath,
+        outputPrefix: this.derivedPagesPrefix(storageRelativePath),
+      })
+      await onPageCountUpdate(pdfPageCount)
+    }
+
+    // Generate temporary URLs for each rendered page of the PDF.
+    return Promise.all(
```
alexisjamet, line 47: **Severity: 🟡 Low** · confirmed by adversarial verification

**`pageCount = 0` is cached forever and yields silent text-only requests.** Only `pdfPageCount === null` triggers rendering, and the converter returns `{pageCount: 0}` with HTTP 200 for a zero-page PDF (the limit check is false and the loop never runs). So 0 is persisted, `Array.from({length: 0})` yields no image parts, and every run on that document 'succeeds' with output hallucinated from the bare prompt — retries hit the cached 0 forever.

**Fix:** treat `pageCount === 0` as an error (converter-side 422 or client-side throw).

apps/api/src/external/llm/ai-sdk-llm-provider-base.ts
```diff
@@ -292,10 +288,6 @@ export abstract class AISDKLLMProviderBase implements LLMProvider {
       ? CallOrigin.streamChatResponse_withTools
       : CallOrigin.streamChatResponse
     this.checkConfigProviderAndModel(config)
```
alexisjamet, line 290: **Severity: 🟡 Low** · confirmed by adversarial verification

**Removing the base-class PDF conversion broke the gated live provider specs.** `ProviderSpecs.testGenerateStructuredOutputFromPdf` still sends a raw `application/pdf` FilePart; with the choke point gone it now throws `MedGemma model cannot process application/pdf file` (MedGemma suite) and inlines the PDF to the image-only Gemma endpoint (Gemma suite). Both suites are env-gated (`GEMMA_TEST`/`MEDGEMMA_TEST`), so CI never sees it — the breakage is silent until the next live run.

**Fix:** update the PDF provider-spec to go through the new page-image flow (or drop it for image-only providers).

apps/api/src/external/llm/providers/medgemma/custom-med-gemma-language-model.ts
```diff
@@ -34,12 +34,11 @@ export class CustomMedGemmaLanguageModel implements LanguageModelV3 {
...
-  get supportedUrls() {
-    return {
-      "image/*": [/^https:\/\/example\.com\/images\/.*/],
-    }
-  }
+  // Image URLs are passed through to the vLLM endpoint, which fetches them
+  // server-side. Restricted to GCS signed urls, the only image urls we
+  // generate (defense in depth against SSRF via injected URLs). Anything else
+  // falls back to the AI SDK downloading the bytes and inlining base64 below.
+  readonly supportedUrls = { "image/*": [/^https:\/\/storage\.googleapis\.com\/.+/] }
```
alexisjamet, line 41: **Severity: 🟡 Low** · confirmed by adversarial verification

**SSRF guard asymmetry: this restriction exists only for MedGemma.** The Gemma provider uses the stock `@ai-sdk/openai` chat model whose `supportedUrls` is `image/* → /^https?:\/\/.*$/` — any http(s) image URL is forwarded verbatim for the Gemma vLLM endpoint to fetch server-side, which is exactly the injected-URL vector this comment says it defends against. Pre-existing exposure (plain image attachments were already URL-borne on Gemma before this PR) and all image URLs are currently constructed server-side from storage paths, so exploiting it needs a separate injection primitive — hence low.

**Fix:** wrap the Gemma model to restrict `supportedUrls` to `https://storage.googleapis.com/` the same way.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  pageObjectPath(storageRelativePath: string, pageNumber: number): string {
+    return `${this.derivedPagesPrefix(storageRelativePath)}page-${pageNumber}.png`
+  }
+
+  async getImageUrls({
+    document: { storageRelativePath, pdfPageCount },
+    onPageCountUpdate,
+    fileStorageService,
```
alexisjamet, line 31: **Severity: 🟡 Low** · confirmed by adversarial verification

**Callers re-wire what this service could own.** The pdf-as-images block is near-verbatim duplicated at both call sites (~30 lines each in `agent-llm-request.service.ts` and `extraction-agent-session-runner.service.ts` — same comments, same structure, already diverging in error mapping), and `fileStorageService` is passed as a runtime parameter although `StorageModule`'s `FILE_STORAGE_SERVICE` could be constructor-injected with no cycle (it imports only `ConfigModule`). The `onPageCountUpdate` callback and the one-site-only entity re-projection add to the per-caller boilerplate.

**Fix:** inject `FILE_STORAGE_SERVICE` into `PdfPagesService`, expose one helper returning ready-made image parts plus `renderedPageCount` (instead of the callback), and let each caller shrink to a few lines.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(private readonly pdfConverterClient: PdfConverterClient) {}
+
+  derivedPagesPrefix(storageRelativePath: string): string {
+    const lastSlashIndex = storageRelativePath.lastIndexOf("/")
+    const directory = storageRelativePath.slice(0, lastSlashIndex)
+    const baseName = storageRelativePath.slice(lastSlashIndex + 1).replace(/\.[^.]+$/, "")
+    return `${directory}/derived/${baseName}/`
+  }
+
+  pageObjectPath(storageRelativePath: string, pageNumber: number): string {
```
alexisjamet, line 24: **Severity: 🟡 Low** · confirmed by adversarial verification

**The `page-{n}.png` naming contract lives independently in two codebases.** This reconstructs object names by convention while the Go converter embeds `fmt.Sprintf("%spage-%d.png", ...)` (`server.go`); only the prefix travels over the wire and no test spans both services. A rename on one side (zero-padding, webp) breaks nothing at render time — the count returns and is cached — and only fails at vLLM fetch time in production, persisting because of the cache.

**Fix:** have `/render-document` return the object names it wrote and sign those, making the converter the single owner of the layout.

Makefile
```diff
@@ -71,9 +71,8 @@ trivy-scan: docker-build
 	trivy image --ignore-unfixed --pkg-types os,library --severity CRITICAL,HIGH --ignorefile .trivyignore.yaml ${localApiImage}
```
alexisjamet, line 71: **Severity: 🟡 Low** · confirmed by adversarial verification

**Local gates silently dropped the PDF service.** This PR removed the pdf image from `trivy-scan`/`docker-build` and the pdf-renderer tests from `tests`/`tests-parallel`/`tests-only-parallel` without wiring in pdf-converter equivalents — `docker-build-pdf-converter` exists but belongs to no aggregate target, and no `go test` appears anywhere in the Makefile. CI compensates for trivy (`security.yml` builds and scans the image), but `pdf-converter-ci.yml` is path-filtered to `apps/pdf-converter/**`, so API-side contract changes never trigger the Go tests.

**Fix:** add the converter image to `trivy-scan`/`docker-build` and a `go test ./...` step to the `tests*` targets.

apps/pdf-converter/main.go
```diff
@@ -0,0 +1,47 @@
...
+		log.Fatalf("gcs client: %v", err)
+	}
+	renderer, err := render.NewRenderer()
+	if err != nil {
+		log.Fatalf("renderer: %v", err)
+	}
+
+	server := newServer(&gcsStore{bucket: client.Bucket(bucketName)}, renderer, maxSourceBytes)
+	log.Printf("pdf-converter listening on :%s (bucket %s)", port, bucketName)
+	log.Fatal(http.ListenAndServe(":"+port, server))
```
alexisjamet, line 59: **Severity: 🟡 Low** · confirmed by adversarial verification

**No server timeouts and no request-body cap.** Plain `http.ListenAndServe` means no `ReadHeaderTimeout`/`ReadTimeout`/`IdleTimeout`, and both handlers `json.NewDecoder(request.Body)` without `http.MaxBytesReader`. Low behind Cloud Run invoker IAM and Google's front end, but a stalled authenticated caller pins a goroutine and its up-to-50MB buffer with no server-side deadline, competing for memory with the render pool.

**Fix:** construct an `http.Server` with timeouts sized to the API client's 120s budget and wrap bodies in `MaxBytesReader`.

apps/pdf-converter/main.go
```diff
@@ -0,0 +1,47 @@
...
+	if port == "" {
+		port = "3002"
+	}
+	maxSourceBytes := int64(50 * 1024 * 1024)
+	if fromEnv := os.Getenv("PDF_CONVERTER_MAX_PDF_BYTES"); fromEnv != "" {
+		parsed, err := strconv.ParseInt(fromEnv, 10, 64)
+		if err != nil {
+			log.Fatalf("invalid PDF_CONVERTER_MAX_PDF_BYTES: %v", err)
+		}
+		maxSourceBytes = parsed
```
alexisjamet, line 33: **Severity: 🟡 Low** · confirmed by adversarial verification

**`PDF_CONVERTER_MAX_PDF_BYTES` accepts 0/negative.** Only the parse error is checked, so a 0 or negative value (or an accidental `50` meaning MB) starts the service cleanly and then every non-empty PDF fails 413 with the confusing `source pdf is 0MB, max is 0MB`.

**Fix:** `log.Fatal` at startup when the parsed value is `<= 0`.

apps/pdf-converter/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+		pdfBytes, ok := fetchSourcePdf(response, request, store, body.SourceObject, maxSourceBytes)
+		if !ok {
+			return
+		}
+
+		pageCount, err := renderer.RenderPages(pdfBytes, body.MaxPages, body.MaxPixelsPerPage,
+			func(pageNumber int, pngBytes []byte) error {
+				object := fmt.Sprintf("%spage-%d.png", body.OutputPrefix, pageNumber)
+				return store.Upload(request.Context(), object, "image/png", pngBytes)
```
alexisjamet: **Severity: 🟡 Low** · confirmed by adversarial verification

**Cancelled render leaves orphaned partial page sets.** Uploads run on `request.Context()`, so a client disconnect/timeout mid-render cancels partway; already-uploaded `page-{n}.png` objects remain under the prefix with no cleanup and no tracking (the object store interface has no Delete). A retry overwrites them, but an abandoned run leaves partial sensitive content invisible to any future cleanup — compounding the broader no-deletion-path issue (see PR comment on derived-page retention).

apps/api/src/external/llm/agent-provider.ts
```diff
@@ -1,5 +1,11 @@
-import { AgentModel } from "@caseai-connect/api-contracts"
+import { AgentModel, AgentModelToAgentProvider, AgentProvider } from "@caseai-connect/api-contracts"
 
 export function GetAgentModelKeyFromValue(model: string) {
   return Object.keys(AgentModel).find((key) => AgentModel[key as keyof typeof AgentModel] === model)
 }
+
+/** Gemma and MedGemma are image-only models: pdfs must be sent as images. */
+export const modelRequiresPdfAsImages = (model: AgentModel | string): boolean => {
```
alexisjamet, line 8: **Severity: 🟡 Low** · confirmed by adversarial verification

**Model file-type constraints are now enforced at three altitudes.** `modelRequiresPdfAsImages` is consulted at the two call sites; the Mistral no-pdf guard lives only inside `generateStructuredOutput` (a pdf chat attachment on Mistral flows unguarded end-to-end — the AI SDK downloads and inlines it; pre-existing); MedGemma's rejection lives inside its custom model. With no per-provider capability abstraction, the next document-attaching feature (e.g. resource-library files, which already have `storageRelativePath` and signed downloads) must simply remember to call the right check — a miss is a silent runtime failure with no type error and no test failure.

**Fix:** a capability on the provider interface (e.g. `documentInputMode`, next to the existing `supportedUrls`) as the single authority.

apps/pdf-converter/internal/render/render.go
```diff
@@ -0,0 +1,137 @@
...
+// PDF points are 1/72"; scale 2 renders ~144dpi so extracted text stays
+// legible.
+const renderScale = 2.0
+
+type Renderer struct {
+	pool pdfium.Pool
+}
+
+func NewRenderer() (*Renderer, error) {
+	pool, err := webassembly.Init(webassembly.Config{
```
alexisjamet, line 37: **Severity: 🟡 Low** · confirmed by adversarial verification

**`MinIdle`/`MaxIdle` are dead config: workers are never reused.** go-pdfium v1.19.8's webassembly backend defaults `ReuseWorkers` to false, so `instance.Close()` calls `InvalidateObject` (not `ReturnObject`) — verified in the module source — and no worker is ever kept idle; every request pays full wasm module instantiation, twice per first render given the `/page-count` pre-flight.

**Fix:** set `ReuseWorkers: true` so `MaxIdle` means something, or drop the idle knobs to stop implying warm reuse.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  private async buildAuthHeaders(converterUrl: string): Promise<Record<string, string>> {
+    if (process.env.PDF_CONVERTER_AUTH !== GOOGLE_IAM_AUTH_MODE) {
+      return {}
+    }
+    // The audience must be the Cloud Run service root URL, not the full path.
+    const audience = new URL(converterUrl).origin
+    this.googleAuth ??= new GoogleAuth()
+    const idTokenClient = await this.googleAuth.getIdTokenClient(audience)
+    const idToken = await idTokenClient.idTokenProvider.fetchIdToken(audience)
```
alexisjamet, line 108: **Severity: 🟡 Low** · confirmed by adversarial verification

**A fresh Google ID token is minted on every converter call.** `getIdTokenClient` returns a new `IdTokenClient` per call, and calling `idTokenProvider.fetchIdToken` directly bypasses the only cache (`IdTokenClient.getRequestMetadataAsync`, which reuses the token until expiry ~1h). Two mints per first render with the pre-flight. (The `GoogleAuth` instance itself is correctly cached via `??=`.)

**Fix:** cache the `IdTokenClient` per audience and use `getRequestHeaders()`.

apps/api/src/domains/documents/storage/gcs-storage.service.ts
alexisjamet, line 54: **Severity: 🟡 Low** · confirmed by adversarial verification

**Every use re-signs every page — 20 `signBlob` API calls per turn in production.** `getImageUrls` signs all pages on every call, including the fully-cached path, and production signing is keyless: each V4 `getSignedUrl` is an `iamcredentials.signBlob` HTTP call (verified in google-auth-library — local crypto only when a key file is loaded). A 20-page document costs 20 signBlob calls per chat turn or extraction run; `Promise.all` bounds wall-clock but quota and per-call overhead accumulate under load.

**Fix:** a short in-process TTL cache of signed URLs keyed by object path (URLs live 15 min; cache ~10).

apps/pdf-converter/store.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("read %s: %w", object, err)
+	}
+	return data, nil
+}
+
+func (store *gcsStore) Upload(ctx context.Context, object string, contentType string, data []byte) error {
+	writer := store.bucket.Object(object).NewWriter(ctx)
```
alexisjamet, line 54: **Severity: 🟡 Low** · confirmed by adversarial verification

**Upload efficiency: serial round trips, 16MiB chunks, extra stat.** (1) Page uploads run synchronously inside the render loop, so each GCS round trip blocks the next page's render — ~20 serial round trips (~1–3s avoidable wall-clock) per 20-page document while the pdfium instance sits idle. (2) The writer keeps the default 16MiB `ChunkSize` for few-hundred-KB PNGs, allocating a 16MiB staging buffer plus resumable-upload semantics per page; `writer.ChunkSize = 0` single-shots since the full byte slice is in hand. (3) `fetchSourcePdf` issues a separate `Attrs` stat RPC before `NewReader`, whose own attrs would do.

**Fix:** bounded-concurrency uploads (errgroup) fed from the render loop, `ChunkSize = 0`, and size-check via the reader's attrs.

apps/api/src/domains/agents/shared/agent-session-messages/streaming/agent-llm-request.service.ts
```diff
@@ -263,44 +270,56 @@ export class AgentLlmRequestService extends ServiceWithLLM {
...
+                await this.agentMessageAttachmentDocumentsService.updatePdfPageCount({
+                  attachmentDocumentId: attachmentDocument.id,
+                  connectScope,
+                  pdfPageCount,
+                })
+              },
+              fileStorageService: this.fileStorageService,
+            })
+            const content = llmMessage.content as Array<ImagePart>
+            imageUrls.map((url) => content.push({ type: "image", image: new URL(url) }))
```
alexisjamet, line 296: **Severity: 🟡 Low** · confirmed by adversarial verification

**`Array.prototype.map` used purely for side effects** (duplicated at `extraction-agent-session-runner.service.ts:243`): it allocates and discards an array of `push()` return values and signals a transformation that never happens. Biome doesn't currently flag it, but a refactor consuming the 'mapped' result would get running array lengths instead of parts.

**Fix:** `content.push(...imageUrls.map((imageUrl) => ({ type: "image", image: new URL(imageUrl) })))` or `for...of`.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+  const buildFileStorageService = () =>
+    ({
+      getTemporaryUrl: jest.fn((storageRelativePath: string) =>
+        Promise.resolve(`https://storage.example.test/${storageRelativePath}?signature=abc`),
+      ),
+    }) as unknown as IFileStorage
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+    delete process.env.PDF_CONVERTER_URL
```
alexisjamet, line 18: **Severity: 🟡 Low** · confirmed by adversarial verification

**Test hygiene, two items.** (1) `afterEach` unconditionally deletes `process.env.PDF_CONVERTER_URL` it may not have set — nothing else sets it today so impact is zero, but save/restore is the hygienic form and avoids ordering-dependent flakiness if `.env.test` ever provides it. (2) This adds a third ad-hoc `IFileStorage` fake cast `as unknown as IFileStorage` (two near-identical fakes already live in csv-extraction-runs); a shared typed helper under `common/test` would keep the fakes compiler-checked against interface changes.

alexisjamet:
**Severity: 🟠 Medium** · confirmed by adversarial verification · *(anchored outside the diff: [`documents.service.ts:277`](https://github.com/bayesimpact/bayes-platform/blob/952433c8f7a4171015c0070e5725b45f1febfb4a/apps/api/src/domains/documents/documents.service.ts#L277))*

**Rendered page PNGs are never deleted — even when the document is.** `deleteDocument` is soft-delete only, `IFileStorage` has no delete method at all, and nothing in the repo or infra deletes GCS objects or records which derived page PNGs exist. Source-PDF retention pre-exists, but this PR materially widens the surface: up to 20 full-document-content images at predictable paths (`{org}/{proj}/derived/{id}/page-{n}.png`) with no deletion or tracking path.

**Failure scenario:** a user or org-offboarding/GDPR-erasure flow deletes a sensitive PDF — the row is soft-deleted but every rendered page image remains in the bucket indefinitely, and because nothing records their existence, even a future delete-on-document-delete fix would silently miss them.

**Fix direction:** add a delete capability to `IFileStorage` and hook document deletion to remove the source object and the `derived/` prefix (or at minimum a bucket lifecycle rule on `derived/` — but see the inline comment on cache invalidation first: today that would permanently break cached documents).

alexisjamet:
**Severity: 🟡 Low** · confirmed by adversarial verification · *(anchored outside the diff: [`gcs-storage.service.ts:76`](https://github.com/bayesimpact/bayes-platform/blob/952433c8f7a4171015c0070e5725b45f1febfb4a/apps/api/src/domains/documents/storage/gcs-storage.service.ts#L76))*

**Presigned uploads allow overwrite, and rendered pages are never re-derived on content change.** `generateSignedUploadUrl` signs V4 write URLs with no `ifGenerationMatch` condition — inconsistent with the direct `save()` path, which uses `ifGenerationMatch: 0` to forbid overwrite. Combined with the new write-once `pdf_page_count` cache, an object overwritten within the 15-minute URL window leaves image-only models serving stale v1 page images forever while text flows, embeddings, and downloads serve v2. Narrow race today (a fresh UUID per presign), but a trap for any future re-upload/replace feature. The precondition-free signing pre-exists; the stale-derived-pages consequence is new with this PR.

**Fix direction:** add `x-goog-if-generation-match: 0` to the signed upload conditions to match the `save()` semantics.

apps/pdf-converter/internal/render/render.go
alexisjamet: ✅ **Fixed in d0418674.**

- `GetPageCount`/`RenderPages` now run under a per-request deadline: new `PDF_CONVERTER_RENDER_TIMEOUT_MS` (default 60s, validated > 0 at startup), kept below the API client's 120s budget so the converter always aborts first — the client comment is now accurate.
- The wazero runtime sets `WithCloseOnContextDone(true)`, and a watchdog calls `instance.Kill()` when the deadline fires or the caller disconnects: the in-flight wasm execution is interrupted and the worker is invalidated so the pool re-creates the slot. The server maps the new `render.ErrAborted` to a **504**.
- Regression tests: a cancellation landing mid-render returns `ErrAborted` promptly (deterministic, no timing assumptions), and a pool test runs **5 interrupted renders against the 4-instance pool** followed by a successful one — it fails if any interruption leaks its slot. A server-level test asserts the 504 when rendering stalls.

The `/healthz` pool-probe suggestion became unnecessary: instances are now always reclaimed, so the pool cannot end up permanently wedged.

apps/api/src/external/llm/providers/medgemma/custom-med-gemma-language-model.ts
alexisjamet: Fixed in #681. `formatMessages` now preserves `image_url` parts (both signed-URL passthrough and base64-inlined bytes) for the streaming path: user messages containing images are sent as OpenAI-style content parts in original order, while text-only messages keep the existing plain-string shape. Added a spec that asserts the request body actually transmitted by `doStreamWithTools` keeps text and image parts — it fails on the pre-fix code with content flattened to the bare text.

apps/api/src/domains/documents/storage/gcs-storage.service.ts
alexisjamet: Fixed in #683: `getTemporaryUrl` now serves signed read URLs from an in-process TTL cache keyed by object path — signed for 15 min, cached for 10, bounded at 5000 entries, evicted on `deleteFile`. The fully-cached `getImageUrls` path no longer makes any signBlob calls.

apps/api/src/domains/agents/shared/agent-session-messages/streaming/agent-llm-request.service.ts
alexisjamet: Fixed in #687 — both occurrences now build the `ImagePart`s with `map` and spread them into a single `content.push(...)`.

apps/pdf-converter/store.go
alexisjamet: Fixed in #690: page uploads now run in a bounded errgroup (limit 4) so GCS round trips overlap with rendering, `writer.ChunkSize = 0` single-shots page PNGs, and `Download` enforces the size limit via the reader's own attrs instead of a separate `Attrs` stat RPC.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
alexisjamet: Fixed in #691: the `IdTokenClient` is now cached per audience and the auth header goes through `getRequestHeaders()`, so the minted ID token is reused until expiry (~1h) instead of minted per request. One wrinkle: `getRequestHeaders()` returns a fetch `Headers` object in google-auth-library v10, so the `authorization` header is extracted explicitly rather than spread.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
alexisjamet: Fixed in #692: `postToConverter` now rejects 200 responses whose body lacks a non-negative-integer `pageCount`, maps non-json 200 bodies to a descriptive error, and maps timeouts during body read to the same message as fetch timeouts. Each failure mode is covered by a test in `pdf-pages.service.spec.ts`.

apps/pdf-converter/internal/render/render.go
alexisjamet: Fixed in #693. Both dimensions are now floored to 1px — and since `max(1, ...)` alone still lets an extreme aspect-ratio page blow the budget through the other dimension (the 1e8×1pt example would still render ~20M px), the larger dimension is re-capped when the product exceeds `maxPixelsPerPage`. Regression tests cover both failure modes (0×0 whole-document error, single-0 budget bypass).

apps/pdf-converter/store.go
alexisjamet: Fixed in #695. `Size()` and the unpinned re-open are gone: `Download(ctx, object, maxBytes)` now checks `reader.Attrs.Size` from the read response itself (same object generation as the bytes, so no stat-then-download divergence) and additionally wraps the read in `io.LimitReader(reader, maxBytes+1)` with a post-read check, making the cap authoritative even against a misreporting backend. This also drops the double GCS round trip. Added a handler-level 413 test; the deleted-between-stat-and-download test was removed since that window no longer exists.

apps/api/src/domains/documents/pdf-pages/pdf-converter.client.ts
alexisjamet: Fixed in #696: the `/page-count` pre-flight is gone — `generatePdfPageImages` now makes a single `/render-document` call, so a first render costs one download, one pdfium parse and one ID-token mint.

The converter's 422 body now carries the document's `pageCount` (`RenderPages` returns the count alongside `ErrTooManyPages`), and the client maps that response to `PdfPageLimitExceededError`, so extraction runs record `PDF_PAGE_LIMIT_EXCEEDED` with the same user-facing message instead of `EXTRACTION_PROVIDER_ERROR`. The `/page-count` endpoint, the client's `getPageCount` and the renderer's `GetPageCount` are deleted; a 422 without `pageCount` (deploy skew with an older converter) falls back to the previous generic message.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
alexisjamet: Fixed in #698: the `-1` case is now handled explicitly — a bare filename maps to a root-level `derived/{name}/` prefix instead of truncating the last character. Comes with a regression test (`derivedPagesPrefix("doc1.pdf")` → `"derived/doc1/"`), verified failing before the fix.

apps/api/src/domains/documents/pdf-pages/pdf-pages.service.ts
alexisjamet: Fixed in #699: `generatePdfPageImages` now throws a user-facing `PdfHasNoPagesError` when the converter reports 0 pages — before rendering and before the count is persisted — and `getImageUrls` no longer trusts a cached `pdfPageCount` of 0 (it re-checks the converter, so rows persisted before the guard fail loudly instead of silently sending the model zero images).

apps/api/src/domains/documents/documents.service.ts
alexisjamet: Fixed in #700: `updatePdfPageCount` now delegates to `documentConnectRepository.updateOneById` and throws `NotFoundException` when nothing matches, so a scope mismatch is surfaced instead of silently skipping the cache. One correction to the comment: `Document` has no soft-delete column, so the `deleted_at` part didn't apply. Covered by a new `service-tests/update-pdf-page-count.spec.ts` (happy path, unknown id, cross-project scope mismatch).

apps/api/src/external/llm/ai-sdk-llm-provider-base.ts
alexisjamet: Fixed in #701: `testGenerateStructuredOutputFromPdf` now branches on `modelRequiresPdfAsImages(model)` — Gemma/MedGemma receive a pre-rendered page image (new checked-in fixture `test-pdf-page-1.png`, rendered with pdfium at scale 2 like the pdf-converter output), mirroring the production page-image flow, while other providers keep the raw pdf FilePart. Assertions unchanged. Note the fix is verified offline only (typecheck + full non-gated suite); the gated `GEMMA_TEST` / `MEDGEMMA_TEST` runs still need to confirm it live.

PR #710, opened 2026-09-02
feat(mcp): OAuth authorization for MCP servers

alexisjamet:
## Summary

MCP servers can now be connected via OAuth 2.1, alongside API keys, following the MCP Authorization spec.

When adding a server, an **Authentication** section offers three choices:

- **API key** — a key field (now required when this method is selected).
- **OAuth** — no key; the user clicks **Authorize** on the server card after creation to run the consent flow.
- **None** — an unauthenticated server.

The chosen method is persisted, so the card reflects it: an API-key or None server shows no button, an OAuth server awaiting its first authorization shows **Authorize**, and a connected one shows **Connected via OAuth**.

### OAuth flow (API-driven)

- Discovery (RFC 9728 / 8414), dynamic client registration (RFC 7591), authorization code + PKCE (S256), token exchange and refresh — all driven by the API.
- All OAuth state (client id, endpoints, tokens, pending authorization with a 10-minute TTL) and the chosen auth method live in the existing encrypted config blob. No DB migration.
- Token refresh happens at connect time under a pessimistic row lock; refresh tokens rotate safely across workers. Tokens are cleared only on a definitive 4xx from the token endpoint; transient network/5xx failures keep them.
- Two endpoints: `POST .../mcp-servers/:id/oauth/initiate` and `POST .../oauth/complete`, both policy-checked. The browser callback lands on the frontend route `/oauth/mcp/callback`, which posts code + state back over the authenticated API. Auth0's SDK is told to skip that route so it doesn't consume the provider's `code`/`state`.
- Also fixes a pre-existing bug where static `headers` were silently dropped when creating an MCP server.

### Security hardening (from the branch security review)

- Discovered OAuth endpoints are validated to be https (http allowed for localhost only), so a hostile server cannot inject a `javascript:` authorization URL into `window.location.assign`.
- The `resource_metadata` URL from `WWW-Authenticate` is only followed when it shares the MCP URL's origin; all OAuth fetches use `redirect: "manual"` and a 10s timeout.
- Tokens are never returned in DTOs, logged, or exposed to the agent tool layer.

## Deployment note

Set `MCP_OAUTH_REDIRECT_URL` (the frontend callback URL, e.g. `https://<app>/oauth/mcp/callback`) in each environment before enabling — see infra PR bayesimpact/infra#181. It is read lazily: if missing, the first Authorize click returns a 500.

## Test plan

- Full API suite green, typecheck, biome, boundaries all clean.
- Coverage: PKCE helpers, discovery/DCR (incl. scheme validation and origin checks), initiate/complete/refresh service paths, e2e on both OAuth endpoints, and e2e on each authentication method at creation (apiKey requires a key, oauth reports pending, none drops any key).
- Manually verified end to end against a live OAuth MCP server (Coros) and a plain no-auth server running side by side.


apps/api/src/domains/mcp-servers/mcp-servers.service.ts
```diff
@@ -110,8 +112,15 @@ export class McpServersService {
...
   }
 
-  decryptUrl(mcpServer: McpServer): string {
-    return this.decryptConfig(mcpServer).url
+  getConfig(mcpServer: McpServer): McpServerConfig {
+    return this.decryptConfig(mcpServer)
+  }
+
+  getAuthStatus(config: McpServerConfig): McpServerAuthStatus {
+    if (config.oauth?.tokens) return "oauthConnected"
```
alexisjamet: > **Priority: P1 (blocker), fixed in cf196feb.** An OAuth server could reach a state where the UI says "Connected", every tool call 401s, and the only recovery is deleting the server.

**Bug: an expired token with no refresh token stays "Connected" forever**

`getAuthStatus` returns `oauthConnected` as soon as tokens exist. But `getValidAccessToken` (mcp-oauth.service.ts:156) returns `null` once the access token has expired and there is no `refreshToken`, and it never clears the stored tokens. So the UI shows "Connected" with no Authorize button (McpServerItem only renders it for `oauthPending`), while the agent connects with `apiKey: undefined` and every tool call 401s. The only recovery is deleting the server.

This is made worse by `expires_in ?? 3600` in `requestTokens` (line 228): a token issued without `expires_in` (allowed by RFC 6749) is treated as expiring after one hour even if it never expires.

Suggested fix: report `oauthPending` when the token is expired and there is no refresh token (or clear the tokens in `getValidAccessToken` on that branch), and stop defaulting a missing `expires_in`.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+    const body = (await response.json().catch(() => ({}))) as {
+      access_token?: string
+      refresh_token?: string
+      expires_in?: number
+      error?: string
+    }
+    if (!response.ok || !body.access_token) {
+      this.logger.warn(`MCP OAuth token request failed: ${response.status} ${body.error ?? ""}`)
+      const isDefinitive = response.status >= 400 && response.status < 500
```
alexisjamet: > **Priority: P1 (blocker), fixed in eb573bf5.** Refresh runs on every agent turn, so one 429 from the authorization server permanently disconnects the server and forces every user through the browser flow again.

**Bug: any 4xx during refresh wipes the stored tokens, including 429**

`isDefinitive` is a plain `400 <= status < 500` check. Refresh runs on every agent turn for every enabled OAuth server with no caching, so a rate-limited authorization server answering 429 (or a 408, or a 4xx from a WAF in front of it) throws `OauthTokenRequestError(isDefinitive = true)`, and `getValidAccessToken` (lines 178-186) persists `tokens: undefined`. A perfectly valid refresh token is discarded and every user has to redo the browser flow.

The specs cover 400 `invalid_grant`, 500 and network error only.

Suggested fix: treat the response as definitive only when `body.error` is a grant-fatal code (`invalid_grant`, `invalid_client`, `unauthorized_client`, `invalid_scope`), or at minimum exclude 408 and 429 from the range.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+    })
+
+    const { pendingAuth: _discarded, ...oauthRest } = config.oauth
+    await this.saveConfig(mcpServer.id, { ...config, oauth: { ...oauthRest, tokens } })
+    return this.mcpServerRepository.findOneByOrFail({ id: mcpServer.id })
+  }
+
+  async getValidAccessToken(mcpServerId: string): Promise<string | null> {
+    // Row lock: refresh tokens rotate, so two workers must not refresh at once.
+    return this.dataSource.transaction(async (manager) => {
```
alexisjamet, line 191: > **Priority: P1 (blocker), fixed in 79bc1dee.** A slow authorization server can starve the whole API's connection pool, so the blast radius is every request, not just MCP.

**Performance: row lock held across a 10s network call on the agent hot path**

`getEnabledServersForAgent` is called on every streamed turn (tools.service.ts:176 via `buildLLMRequest`), on every sub-agent invocation (sub-agent-tools.ts:215) and on every messages GET that contains an MCP App (mcp-app-html.service.ts:114). For each OAuth server, in parallel via `Promise.all`, it now runs `BEGIN` + `SELECT ... FOR UPDATE` + `COMMIT` and decrypts a config the caller already decrypted at mcp-servers.service.ts:41. The fresh-token early return at line 153 sits inside the transaction, so there is no lock-free fast path.

During a refresh, `requestTokens` (10s timeout) runs while the row lock is held. Every concurrent turn on that server then blocks on the lock, each holding one of the default 10 pg connections (typeorm.ts sets no `max`). A slow authorization server can starve the pool for the whole API.

Suggested fix: check `expiresAt` on the config the caller already decrypted and only enter the locking transaction when a refresh is due, re-checking inside the lock.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth-discovery.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export async function discoverOauthConfiguration(
+  mcpUrl: string,
+): Promise<McpOauthDiscovery | null> {
+  const resourceMetadataUrl = await probeForResourceMetadataUrl(mcpUrl)
+  const resourceMetadata = await fetchJson<ProtectedResourceMetadata>(resourceMetadataUrl)
+  if (!resourceMetadata?.authorization_servers?.length) return null
+
+  const issuer = (resourceMetadata.authorization_servers as string[])[0]!.replace(/\/$/, "")
+  const serverMetadata =
```
alexisjamet: > **Priority: P2 (should fix).** OAuth is simply unavailable for a whole class of servers, with a misleading error message, though the OIDC fallback covers the common providers.

**Bug: the RFC 8414 path-insertion metadata URL is never tried**

For an issuer with a path (`https://auth.example.com/tenant-a`), the code fetches `.../tenant-a/.well-known/oauth-authorization-server` then `.../tenant-a/.well-known/openid-configuration`. RFC 8414 and the MCP spec put the path-insertion form first: `https://auth.example.com/.well-known/oauth-authorization-server/tenant-a`. An OAuth-only server that publishes just that form (no OIDC endpoint) 404s on both attempts, `discoverOauthConfiguration` returns `null`, and the user sees "This MCP server does not advertise OAuth authorization" although it does.

Keycloak and similar still work through the OIDC fallback, so the impact is limited to pure OAuth path issuers. `mcp-oauth-discovery.spec.ts` only covers a path-less issuer; a test with a path issuer would catch this.

alexisjamet (commented):
Automated review of this PR (Claude Code, high effort): 8 finder angles, then each candidate adversarially verified against the branch. 21 findings survived, posted as inline comments, most severe first:

**Bugs**: expired token without refresh token stuck on "Connected"; any 4xx on refresh wipes tokens; RFC 8414 path-insertion discovery URL never tried; expired Auth0 session on callback drops code/state; initiate on an apiKey server drops the key; unlocked config write can reinstate a rotated refresh token; single localStorage key breaks concurrent flows; unauthorized OAuth servers dialed unauthenticated every turn.

**Performance**: row lock held across the token-endpoint fetch on every agent turn.

**Tests**: no role-based coverage for the two new `canCreate` routes.

**Cleanups**: duplicated URL allowlist, config codec, timeout constant, DTO mapping, spec fixtures and payload type; redundant casts; dead-end error view; extra SELECT after completion.

Rejected during verification (not posted): trailing-slash mismatch on `skipRedirectCallback`, test-transaction visibility of `dataSource.transaction`, unused `completeMcpServerOauth` payload, localStorage helper reuse with the invitation flow.

apps/web/src/common/routes/Router.tsx
```diff
@@ -32,6 +33,7 @@ const router = () =>
       ),
       children: [
         onboardingRoute,
+        mcpOauthCallbackRoute,
```
alexisjamet, line 36: > **Priority: P2 (should fix).** Silent failure of the authorization flow with no error shown, though the user can recover by clicking Authorize again.

**Bug: an expired Auth0 session on the way back drops the OAuth code and state**

This callback route sits under `ProtectedRoute`, which calls `loginWithRedirect` with no `appState.returnTo`, and `auth0.config.ts` sets `redirect_uri` to the origin. When the user returns from the provider's consent screen and Auth0 cannot silently restore the session (expired or revoked refresh token, cleared storage), `ProtectedRoute` sends them to Auth0 login, Auth0 sends them back to `/`, `HomeRoute` navigates to onboarding, and the single-use `code` and `state` are gone. No error is shown, and the pending context stays in localStorage for a later callback to pick up.

Suggested fix: pass `appState: { returnTo: pathname + search }` in `ProtectedRoute` and provide an `onRedirectCallback` on `Auth0Provider` that navigates to `appState.returnTo` (the default only does `history.replaceState`, which react-router does not see).

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
```diff
@@ -38,12 +35,17 @@ export class McpServersService {
...
-      }))
+    const enabledServers = agentMcpServers.filter((agentMcpServer) => agentMcpServer.mcpServer)
+    return Promise.all(
+      enabledServers.map(async (agentMcpServer) => {
+        const { oauth, ...config } = this.decryptConfig(agentMcpServer.mcpServer)
+        if (!oauth) return { id: agentMcpServer.mcpServer.id, ...config }
+        const accessToken = await this.mcpOauthService.getValidAccessToken(
+          agentMcpServer.mcpServer.id,
+        )
+        return { id: agentMcpServer.mcpServer.id, ...config, apiKey: accessToken ?? undefined }
```
alexisjamet: > **Priority: P2 (should fix).** Unrecoverable without deleting the server, but only reachable through the API since the UI never offers Authorize on an apiKey server.

**Bug: calling initiate on an apiKey server silently drops its API key**

The initiate endpoint checks only `canCreate`, not `authMethod`. On a server created with `authMethod: "apiKey"`, one `POST .../oauth/initiate` stores `oauth.pendingAuth` next to the key. From then on `getAuthStatus` returns `oauthPending` (line 121 wins over the apiKey branch at 122), and this line replaces the spread `apiKey` with `accessToken ?? undefined` because there are no tokens yet. Every agent turn dials the server with no Authorization header, and there is no update endpoint to switch back.

The shipped UI never shows Authorize on an apiKey server, so this is API-reachable only, but it is one request away.

Suggested fix: reject initiate when `config.authMethod === "apiKey"`, or make initiate explicitly switch `authMethod` to `oauth` and clear `apiKey`. Note that the e2e `createServer` posts without `authMethod` (inferred `none`) and initiates on it, so the guard should reject `apiKey` rather than require `oauth`.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+      scope,
+      tokens: config.oauth?.tokens,
+      pendingAuth: {
+        state,
+        codeVerifier,
+        redirectUri,
+        expiresAt: Date.now() + PENDING_AUTH_TTL_MS,
+      },
+    }
+    await this.saveConfig(mcpServer.id, { ...config, oauth })
```
alexisjamet: > **Priority: P3 (nice to have).** Needs a re-initiation to overlap with a token refresh, and the worst case is one extra trip through the consent flow.

**Race: unlocked whole-config write can reinstate a rotated refresh token**

`config` is read at line 50 from the entity the guard loaded, then discovery and registration await up to ~40s of fetches, then this line writes the whole blob back through a plain `update` with no lock or version check. Meanwhile a conversation turn can refresh under the row lock in `getValidAccessToken` and persist a rotated refresh token (line 169). This write then puts the old, now revoked, refresh token back (`tokens: config.oauth?.tokens` at line 83). If the user abandons the consent page, the next refresh gets `invalid_grant` and the tokens are wiped.

It needs a re-initiation on an already connected server plus an unlucky overlap with token expiry, so it is not frequent, but the e2e spec shows re-initiation is a supported path. `completeAuthorization` (line 137) has the same shape but writes fresh tokens, so the overwrite there is harmless.

Suggested fix: re-read and patch only the `oauth` sub-object inside the same pessimistic-lock transaction, rather than persisting the pre-network snapshot.

apps/web/src/studio/features/mcp-servers/mcp-oauth-storage.ts
```diff
@@ -0,0 +1,33 @@
+const STORAGE_KEY = "mcpOauthPendingContext"
```
alexisjamet: **Bug: a single localStorage key breaks two concurrent Authorize flows**

Two Authorize clicks in two tabs write the same key, last writer wins. When provider A returns first, the callback takes context B and posts to server B's complete route with state A, the API answers "OAuth state mismatch", and when provider B returns the key is already empty so the user sees `missingContext`. Both flows fail and must be redone. It is fail-safe (state is checked before any token exchange), so severity is low.

Since `state` only lives inside the encrypted config blob, keying the API lookup by state is not cheap. The frontend-only fix is easy though: the authorization URL returned by the API already carries `state` as a query param, so store the context under `mcpOauthPendingContext:<state>` and read it by the `state` received on callback.

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
```diff
@@ -38,12 +35,17 @@ export class McpServersService {
...
-      .map((agentMcpServer) => ({
-        id: agentMcpServer.mcpServer.id,
-        ...this.decryptConfig(agentMcpServer.mcpServer),
-      }))
+    const enabledServers = agentMcpServers.filter((agentMcpServer) => agentMcpServer.mcpServer)
+    return Promise.all(
+      enabledServers.map(async (agentMcpServer) => {
+        const { oauth, ...config } = this.decryptConfig(agentMcpServer.mcpServer)
+        if (!oauth) return { id: agentMcpServer.mcpServer.id, ...config }
+        const accessToken = await this.mcpOauthService.getValidAccessToken(
```
alexisjamet: > **Priority: P2 (should fix).** This is the normal state of a freshly created OAuth server, so the agent quietly loses that server's tools with no signal to anyone.

**UX/observability: an unauthorized OAuth server is dialed unauthenticated on every turn**

A server created with `authMethod: "oauth"` that was never authorized (or whose tokens were dropped after a definitive refresh failure) is still returned to the agent here with `apiKey: undefined`. `McpClientService.connect` then dials it without Authorization on every turn, catches the 401, logs an error and returns `tools: {}`. The agent silently has no tools from that server and the end user gets no signal.

The silent catch predates this PR (a wrong API key already behaved this way), but the PR turns it into a designed, common state. Suggested fix: skip servers whose OAuth state has no usable token (with a warn naming the server) instead of dialing unauthenticated, and consider surfacing a "needs authorization" hint where the server is enabled on an agent.

apps/api/src/domains/mcp-servers/e2e-tests/oauth.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  it("rejects completion with a bad state", async () => {
+    await createContext()
+    mcpServerId = await createServer()
+
+    const response = await complete({ payload: { code: "code-1", state: "wrong" } })
+
+    expectResponse(response, 400)
+  })
+
+  it("requires authentication on both routes", async () => {
```
alexisjamet, line 204: > **Priority: P2 (should fix).** apps/api/CLAUDE.md requires this coverage, and without it a policy regression on the two new routes ships green.

**Tests: the two new `canCreate` routes have no role-based coverage**

apps/api/CLAUDE.md asks for an `auth.spec.ts` per domain covering no token, not a member, wrong role and allowed roles for every route, using `createContextForRole`. The mcp-servers domain has no `auth.spec.ts` (pre-existing gap), and the only authorization check for `initiateOauth` and `completeOauth` is this no-token case in a functional spec. A plain member hitting either route is untested, so a policy regression would ship green.

Minor on top: this test runs `createContext()` plus an authenticated `createServer()` only to get an id before asserting 401, which the guard returns before resolving any context. `agents/e2e-tests/auth.spec.ts` uses `randomUUID()` for this.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+    private readonly mcpServerRepository: Repository<McpServer>,
+    private readonly encryptionService: EncryptionService,
+    private readonly configService: ConfigService,
+    @InjectDataSource()
+    private readonly dataSource: DataSource,
+  ) {}
+
+  async initiateAuthorization(mcpServer: McpServer): Promise<{ authorizationUrl: string }> {
+    const config = this.readConfig(mcpServer)
+    const redirectUri = this.configService.getOrThrow<string>("MCP_OAUTH_REDIRECT_URL")
```
alexisjamet: > **Priority: P3 (nice to have).** Consistent with how the codebase reads other variables, and the deployment note already calls it out.

**Nit: `MCP_OAUTH_REDIRECT_URL` is only validated on the first Authorize click**

Read lazily with `getOrThrow`, so a deployment that forgets the new variable boots fine and returns a 500 on the first click, surfaced only as the generic "MCP server authorization failed" toast. The smoke compose sets `MCP_ENCRYPTION_KEY` but not this one.

This matches how the `AUTH0_*` variables are read, so it is consistent with the codebase. Reading it once in the constructor would make a misconfigured deploy fail at boot, the way `EncryptionService` does.

apps/web/src/studio/features/mcp-servers/mcp-servers.thunks.ts
```diff
@@ -1,11 +1,27 @@
...
 import { getApiErrorMessage } from "@/common/utils/api-error"
+import { savePendingMcpOauthContext } from "./mcp-oauth-storage"
 import type { McpServer } from "./mcp-servers.models"
 
 type ThunkConfig = { state: RootState; extra: ThunkExtraArg; rejectValue: string }
 
+// Defense in depth: the API already validates discovered OAuth endpoints, but
+// this is the last line before the URL reaches window.location.assign, a
+// javascript:/data: URL there would execute on the app origin.
+const isSafeAuthorizationUrl = (candidate: string): boolean => {
```
alexisjamet: > **Priority: P2 (should fix).** A security allowlist duplicated across two packages will drift, and the copy that drifts is the one guarding a browser redirect.

**Reuse: `isSafeAuthorizationUrl` is byte-identical to `isValidEndpointUrl`**

Same body as `isValidEndpointUrl` in `apps/api/.../oauth/mcp-oauth-discovery.ts:28`. This is a security-relevant https-or-localhost allowlist now maintained in two packages, and nothing ties them together. `api-contracts` already ships runtime helpers (`defineRoute`, `buildNameFromEmail`) and is a dependency of both apps, so one exported helper there would do.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+      )
+    }
+    return {
+      accessToken: body.access_token,
+      refreshToken: body.refresh_token,
+      expiresAt: Date.now() + (body.expires_in ?? 3600) * 1000,
+    }
+  }
+
+  private readConfig(mcpServer: McpServer): McpServerConfig {
```
alexisjamet: **Reuse: the config encrypt/decrypt codec is hand-rolled in five places**

`readConfig` and `persistConfigWithManager` re-implement what `McpServersService.decryptConfig` (line 126) and its two `encrypt(JSON.stringify(config))` calls (lines 52 and 72) already do. Any future change to the blob (schema version, zod parse, migration shim) has to be replicated in each copy.

Both services already inject `EncryptionService`, and `McpServersService` now depends on `McpOauthService`, so the codec belongs on `EncryptionService` (`encryptConfig` / `decryptConfig`) rather than on either service.

apps/api/src/domains/mcp-servers/mcp-servers.controller.ts
```diff
@@ -94,14 +113,46 @@ export class McpServersController {
...
+      code: payload.code,
+      state: payload.state,
+    })
+    const config = this.mcpServersService.getConfig(updated)
+    return { data: toMcpServerDto(updated, config, this.mcpServersService.getAuthStatus(config)) }
+  }
 }
 
-function toMcpServerDto(entity: McpServer, url: string): McpServerDto {
+function toMcpServerDto(
```
alexisjamet: > **Priority: P3 (nice to have).** Cleanup.

**Simplification: the getConfig, getAuthStatus, toMcpServerDto triple is repeated three times**

`toMcpServerDto` takes `authStatus` as a parameter although `getAuthStatus` is a pure function of `config`, so lines 67, 79-80 and 140-141 each repeat the same three steps. The next endpoint returning a server has to remember the dance or its `authStatus` will disagree with the list.

Either let `toMcpServerDto(entity, config)` compute the status itself, or add a `toDto(entity)` on the service that owns the whole mapping.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.spec.ts
```diff
@@ -0,0 +1,452 @@
...
+import { agentFactory } from "@/domains/agents/agent.factory"
+import { createOrganizationWithProject } from "@/domains/organizations/organization.factory"
+import { EncryptionService } from "../encryption.service"
+import type { McpServer } from "../mcp-server.entity"
+import { McpServersModule } from "../mcp-servers.module"
+import type { McpServerOauthState, McpServerOauthTokens } from "../mcp-servers.service"
+import { McpServersService } from "../mcp-servers.service"
+import { McpOauthService } from "./mcp-oauth.service"
+
+global.fetch = jest.fn()
```
alexisjamet: **Simplification: 41 fixture lines are duplicated between this spec and the e2e spec**

Lines 19-59 (fetch mock, metadata fixtures, `json`, `unauthorizedProbe`, `mockFullDiscoveryAndRegistration`) are identical to `e2e-tests/oauth.spec.ts:19-59`. A change to the discovery flow has to be mirrored in both mock chains, and updating only one leaves a green suite that no longer exercises the same protocol.

Also, all five e2e tests start with `await createContext(); mcpServerId = await createServer()`, which could move into the existing `beforeEach`.

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
```diff
@@ -1,26 +1,22 @@
...
-   * encrypted config blob, so adding them needs no migration. The conversation
-   * context is applied after them and cannot be overridden here.
-   */
-  headers?: Record<string, string>
-}
-
-export type EnabledMcpServer = McpServerConfig & {
-  id: string
-}
+export type {
```
alexisjamet: **Simplification: type re-exports plus a `getConfig` wrapper add indirection**

These five types are re-exported from `./mcp-server-config.types` while `mcp-oauth.service.ts` imports them from the types file directly, so there are two import paths for the same types. And `getConfig` (line 115) is a one-line public wrapper around the private `decryptConfig` (line 126), so readers have to check whether they differ.

Pointing the controller and the oauth spec at the types file, dropping the re-export block and keeping a single public method would tidy this.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth-discovery.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export async function discoverOauthConfiguration(
+  mcpUrl: string,
+): Promise<McpOauthDiscovery | null> {
+  const resourceMetadataUrl = await probeForResourceMetadataUrl(mcpUrl)
+  const resourceMetadata = await fetchJson<ProtectedResourceMetadata>(resourceMetadataUrl)
+  if (!resourceMetadata?.authorization_servers?.length) return null
+
+  const issuer = (resourceMetadata.authorization_servers as string[])[0]!.replace(/\/$/, "")
```
alexisjamet: **Simplification: redundant cast, non-null assertion and a dead nullish guard**

`authorization_servers` is already typed `string[] | undefined` and the length check above narrows it, so `as string[]` and `!` only silence the compiler at the exact spot where third-party JSON is consumed. `const [issuerBase] = resourceMetadata.authorization_servers; if (!issuerBase) return null` says the same thing without either.

Related, in `probeForResourceMetadataUrl` (line 147) `new URL(mcpUrl)` is parsed again although line 120 already did, and `url.pathname ?? ""` is dead since `pathname` is always a string.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+import { McpServer } from "../mcp-server.entity"
+import type { McpServerConfig, McpServerOauthState } from "../mcp-server-config.types"
+import { discoverOauthConfiguration, registerOauthClient } from "./mcp-oauth-discovery"
+import { codeChallengeS256, generateCodeVerifier, generateState } from "./pkce"
+
+const PENDING_AUTH_TTL_MS = 10 * 60 * 1000
+const TOKEN_REFRESH_MARGIN_MS = 60 * 1000
+// The refresh path runs inside a pessimistic row lock (see getValidAccessToken
+// below), so a hanging authorization server must not hold that lock forever.
+const OAUTH_FETCH_TIMEOUT_MS = 10_000
```
alexisjamet: **Reuse: `OAUTH_FETCH_TIMEOUT_MS` and the fetch options are declared twice**

Same constant name and value in `mcp-oauth-discovery.ts:19`, and the `redirect: "manual"` plus `AbortSignal.timeout(...)` shape appears four times across the two files. Exporting the constant (or a small `oauthFetchInit()` helper) from the discovery module and importing it here would keep the two timeouts from drifting.

apps/web/src/studio/features/mcp-servers/mcp-servers.spi.ts
```diff
@@ -1,14 +1,25 @@
+import type { McpServerAuthMethod } from "@caseai-connect/api-contracts"
 import type { McpServer } from "./mcp-servers.models"
 
 type ProjectScope = { organizationId: string; projectId: string }
 type McpServerScope = ProjectScope & { mcpServerId: string }
 type McpServerAgentScope = McpServerScope & { agentId: string }
-type CreateMcpServerFields = { name: string; url: string; apiKey?: string }
+type CreateMcpServerFields = {
```
alexisjamet: **Simplification: the create payload shape is defined four times**

`CreateMcpServerFields` here, then the same four fields inline in `mcp-servers.thunks.ts:40`, `McpServersList.tsx:21-25` and `McpServersRoute.tsx:28-32`, while `api-contracts` already exports `CreateMcpServerDto`. Adding a field to the dialog means touching all four.

Cheapest fix: export this type and use it in the other three places. Switching to `CreateMcpServerDto` also works, with the note that it carries an optional `headers` field the literals do not.

apps/web/src/common/routes/McpOauthCallbackRoute.tsx
```diff
@@ -0,0 +1,60 @@
...
+            organizationId: context.organizationId,
+            projectId: context.projectId,
+          }),
+          { replace: true },
+        ),
+      )
+      .catch(() => setError(t("mcpServers:oauthCallback.failed")))
+  }, [dispatch, navigate, searchParams, t])
+
+  if (error) {
```
alexisjamet: **UX: the error view is a dead end**

This block mirrors the centered layout of `ErrorRoute.tsx` but, unlike it, offers no link or button out. The copy tells the user to start again from the MCP servers page without taking them there, so they have to edit the URL.

Either reuse `ErrorRoute` (accepting its generic title) or add a link to the MCP servers page. A shared layout abstraction beyond that is not worth it for two six-line blocks.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,252 @@
...
+      code,
+      redirect_uri: pendingAuth.redirectUri,
+      client_id: config.oauth.clientId,
+      code_verifier: pendingAuth.codeVerifier,
+      resource: config.oauth.resource,
+    })
+
+    const { pendingAuth: _discarded, ...oauthRest } = config.oauth
+    await this.saveConfig(mcpServer.id, { ...config, oauth: { ...oauthRest, tokens } })
+    return this.mcpServerRepository.findOneByOrFail({ id: mcpServer.id })
```
alexisjamet: **Minor: extra SELECT and second decrypt after completing OAuth**

`completeAuthorization` already holds the final `config` but re-selects the entity here, and the controller (mcp-servers.controller.ts:140) decrypts the blob again just to derive `url` and `authStatus`. Once per authorization, so negligible, but it repeats the double-decrypt pattern from the hot path. Returning the config alongside the entity (or using `UPDATE ... RETURNING` for `updatedAt`) would remove both.

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
alexisjamet: Fixed in cf196feb.

- `expiresAt` is now optional on the stored tokens. A token response without `expires_in` stores no expiry and the token is used until the server rejects it, instead of being treated as expiring after one hour.
- `getAuthStatus` derives `oauthConnected` from whether the tokens can still authenticate (fresh, no known expiry, or refreshable) rather than from their mere presence. An expired token with no refresh token now reports `oauthPending`, so the Authorize button comes back.
- `getValidAccessToken` drops the dead tokens on that branch (with a warn log), mirroring the definitive-refresh-failure path, so the stored config stays consistent with the status.

The two checks live in a small `oauth/oauth-tokens.ts` helper shared by both services. Specs added for the no-expiry token, the expired token without refresh token, the token response without `expires_in`, and the pure `getAuthStatus` cases. Biome, API and web typecheck, and the mcp-servers unit, e2e and discovery suites pass.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth-discovery.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Probes the MCP endpoint expecting a 401 whose WWW-Authenticate names the
+ * resource metadata URL. Falls back to the RFC 9728 path-derived well-known
+ * URL when the header is absent, points off-origin (a malicious or misbehaving
+ * server naming an attacker-controlled metadata host), or the probe itself
+ * fails.
+ */
+async function probeForResourceMetadataUrl(mcpUrl: string): Promise<string> {
+  const mcpOrigin = new URL(mcpUrl).origin
+  try {
+    const probe = await fetch(mcpUrl, {
```
alexisjamet, line 130: > **Priority: P2 (should fix), fine as a follow-up issue.** Real SSRF surface, but it needs a project member with create rights, it is a blind probe, and the same check belongs on the MCP client's own connection.

**Security: discovery fetches are an SSRF surface (flagged by automated security review)**

Discovery POSTs to the user-supplied MCP URL, then GETs the `resource_metadata` URL taken from that server's `WWW-Authenticate` header (or the well-known path), then GETs metadata URLs derived from whatever `authorization_servers` that document names. A project member with create rights can point all of this at internal hosts (metadata endpoints, other services on the VPC, cloud instance metadata at 169.254.169.254). `isValidEndpointUrl` (line 28) only checks the scheme and accepts plain http for `localhost` and `127.0.0.1` in every environment, and it is applied to the returned endpoints, not to the MCP URL or the discovered issuer.

Mitigating factors, so this is a judgment call rather than a blocker: the MCP client already connects to any URL a member enters, so the trust model is not new; the fetched bodies are parsed as JSON and never returned to the caller, so it is a blind probe; and the one URL that reaches the browser (the authorization endpoint) is validated before the redirect.

If we want to tighten it:
- A shared host check that resolves DNS and rejects loopback, `0.0.0.0/8`, RFC 1918 (`10/8`, `172.16/12`, `192.168/16`) and link-local (`169.254/16`, `fe80::/10`), applied to `mcpUrl` before the probe, to the discovered issuer before it is interpolated into the well-known URL, and inside `isValidEndpointUrl`.
- Gate the localhost exception behind an explicit development flag instead of shipping it in production.

Since the same check would apply to the MCP client's own connection, this may be better as a follow-up issue covering both than as part of this PR.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
alexisjamet: Fixed in eb573bf5.

`requestTokens` no longer treats every 4xx as definitive. It now reads the RFC 6749 §5.2 error code from the response body and marks the refresh as definitive only for `invalid_grant`, `invalid_client`, `unauthorized_client` and `invalid_scope`. A 429, a 408, or a 4xx from a WAF with no OAuth body is treated like a network error or 5xx: the refresh returns `null` for this turn and the stored tokens stay untouched for the next attempt.

Specs added for the three cases this comment called out: a 429 keeps the tokens, a 403 with an HTML body keeps the tokens, and a 401 `invalid_client` still clears them. All nine mcp-servers suites, biome and the API typecheck pass.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
alexisjamet: Fixed in 79bc1dee.

`getValidAccessToken` now takes the oauth state the caller already decrypted and answers from it without touching the database while the access token is fresh (or when there are no tokens at all). Only a refresh that is actually due enters the pessimistic-lock transaction, which moved to a private `refreshAccessToken`. Inside the lock the state is still re-read and re-checked, so when another worker refreshed between the caller's lock-free check and lock acquisition, its fresh token is reused instead of triggering a second round trip to the authorization server.

In `getEnabledServersForAgent` this means a fresh-token server costs zero extra queries and zero extra decrypts per turn; the double decrypt only remains on the refresh path, where the re-read under the lock is the point.

Specs: the fast path opens no transaction (asserted on `getValidAccessToken` directly and through `getEnabledServersForAgent`), the no-tokens case opens no transaction, and a stale caller state against an already refreshed row returns the row's token without calling the token endpoint. All nine mcp-servers suites, biome and the API typecheck pass.

Not changed here: the lock is still held across the 10s token endpoint call during the refresh itself. That is now bounded to one refresh per expiry per server rather than every turn, and concurrent turns that arrive mid-refresh wait for the fresh token rather than each firing their own refresh, which is the behaviour we want with rotating refresh tokens.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth.service.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  private readConfig(mcpServer: McpServer): McpServerConfig {
+    return JSON.parse(this.encryptionService.decrypt(mcpServer.encryptedConfig)) as McpServerConfig
+  }
+
+  private async saveConfig(mcpServerId: string, config: McpServerConfig): Promise<void> {
+    await this.persistConfigWithManager(this.dataSource.manager, mcpServerId, config)
+  }
+
+  private async persistConfigWithManager(
```
alexisjamet: @jdoucy please review

apps/api/src/domains/mcp-servers/oauth/mcp-oauth-discovery.ts
alexisjamet: Addressed in #752: discovery now tries the RFC 8414 path-insertion metadata URLs first for issuers with a path.

apps/web/src/common/routes/Router.tsx
alexisjamet: Addressed in #754: ProtectedRoute passes a sanitized returnTo and the Auth0 provider navigates back to the callback after re-login.

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
alexisjamet: Addressed in #755: initiating OAuth on an API key server is rejected with a 400 before any discovery fetch, and the key is preserved.

apps/api/src/domains/mcp-servers/mcp-servers.service.ts
alexisjamet: Addressed in #757: OAuth servers with no usable token are skipped with a warning instead of being dialed unauthenticated. The frontend needs-authorization hint is left as a follow-up.

apps/api/src/domains/mcp-servers/e2e-tests/oauth.spec.ts
alexisjamet: Addressed in #756: the existing auth spec now covers initiateOauth and completeOauth plus the missing not-a-member cases, and the no-token test uses randomUUID().

apps/web/src/studio/features/mcp-servers/mcp-servers.thunks.ts
alexisjamet: Addressed in #753: a single isAllowedOauthEndpointUrl helper in api-contracts replaces both copies.

apps/api/src/domains/mcp-servers/oauth/mcp-oauth-discovery.ts
alexisjamet: Addressed in #758: a DNS-resolving host guard rejects private, loopback and link-local targets on every discovery fetch, with http localhost gated behind MCP_OAUTH_ALLOW_INSECURE_LOCAL. Applying the same guard to the MCP client connection is left as a follow-up.

apps/web/src/common/routes/Router.tsx
alexisjamet: won't fix

alexisjamet: Update: #754 was closed without merging, this issue is intentionally not being pursued.

PR #723, opened 2026-09-03
fix(api): allow any connect.localhost port in local CORS fallback

alexisjamet:
## Summary

Vite dev and preview ports are configurable per checkout (`FRONT_PORT`, `FRONT_PREVIEW_PORT`), and worktrees override them to run side by side. The default CORS origins pinned `https://connect.localhost:5173` and `:5174`, so any checkout off the default ports hit CORS errors until `FRONTEND_URL` was set by hand.

Outside production, the fallback now matches any port on `https://connect.localhost`. Behavior is otherwise unchanged:

- `FRONTEND_URL`, when set, still replaces the fallback entirely.
- `FRONTEND_URL` stays required in production and the API still fails at boot without it.
- Public embed endpoints keep reflecting any origin, everything else stays pinned.

Also updates `.env-example` (the variable is now commented out for local use), the README troubleshooting note, and the changelog.

## Test plan

- `cors.spec.ts` covers the accepted local ports, rejected lookalike hosts, the blank/unset fallback, the production throw, and that a set `FRONTEND_URL` disables the fallback.
- `npx tsc --noEmit` and biome pass on the touched files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 
 ### Changed
 
+- Local development: the API now allows any connect.localhost port without setting FRONTEND_URL.
+
 ### Fixed
 
 ### Security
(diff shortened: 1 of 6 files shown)


alexisjamet:
refused by @jdoucy

PR #731, opened 2026-09-04
fix(llm): strip leaked tool calls with no angle bracket

jdoucy:
## Summary

A new tool-call leak variant was seen in production: `Lie:default_api:mandatory_tool{categoryNames:[],suggestedTitle:null}`. There is no `<` at all, the `default_api:` namespace is glued to a garbled word, and only the argument braces delimit the call. Every existing regexp required a leading `<`, so the text reached the user.

- Add a bare `(prefix:)?default_api:tool{...}` alternative to the cleanup regexp (the glued prefix is removed with the call).
- Streaming stripper now holds the buffer back on an open `default_api:` until its closing `}`.
- `findLeakedToolCalls` extracts the tool name from this variant too, so recovery still applies.
- Changelog entry.

## Test plan

- [x] New specs: full-text cleanup, streaming cleanup split across 7-char deltas, tool-name extraction
- [x] `npx jest` on thought-tokens-helper and leaked-tool-call-recovery specs (20 passing)
- [x] biome check and tsc on apps/api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ### Changed
 
 ### Fixed
+- Replies no longer show leaked internal tool-call text, even without angle brackets.
 - Refreshing the page during a reply no longer loses the answer.
 - An interrupted reply can be sent again.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ### Changed
 
 ### Fixed
+- Replies no longer show leaked internal tool-call text, even without angle brackets.
```
jdoucy: retire

jdoucy: Retirée.

PR #739, opened 2026-09-07
feat(pdf-export): built-in MCP server that turns markdown into a downloadable PDF

alexisjamet:
## Summary

Agents can now hand the user a PDF. A built-in "PDF export" MCP server, backed by the `apps/pdf-converter` Go service, exposes one tool, `convert_markdown_to_pdf`. Typical uses: summarise a conversation as a PDF, or export the answers collected by `fillForm`.

### Go service (`apps/pdf-converter`)
- New stateless streamable-HTTP MCP endpoint at `POST /mcp` (official go-sdk v1.7.0).
- New `internal/mdpdf` package lays out GitHub-flavored markdown with goldmark + fpdf (headings, lists, task lists, tables, code blocks, links, blockquotes, page numbers), with DejaVu fonts embedded (3 MB). Images are text only, raw HTML is dropped, output is capped at 200 pages and 1 MiB of input, four concurrent renders.
- The PDF is uploaded under `tmp/pdf-exports/<uuid>/<name>.pdf` and returned as a 15 minute V4 signed URL (signed before upload, so a signing failure leaves nothing behind). The model's text never contains the link.
- The tool declares an MCP App card (`ui://pdf-export/download-card`): file name, page count, size, live expiry countdown, Download button, expired state. It only accepts `https:` URLs and uses `textContent` everywhere.
- Env: `PDF_EXPORT_TTL_MINUTES` (15, max 10080), `PDF_EXPORT_MAX_MARKDOWN_BYTES` (1 MiB), `PDF_EXPORT_TMP_PREFIX` (`tmp/pdf-exports/`). README documents the endpoint, contract and signing requirements.

### API (`apps/api`, `packages/api-contracts`)
- A built-in server row (`presetSlug = pdf-export`, `projectId = null`) is upserted at API boot from `PDF_CONVERTER_URL` (non-fatal on failure). Only allowlisted slugs count as built-in, so operator-seeded presets keep today's behaviour.
- `McpServerDto` gains `isBuiltIn` and a nullable `projectId`. Built-ins are listed first in every project, can be toggled per agent, and deleting one returns 403.
- New shared `external/google-iam` mints Google ID tokens; the MCP client uses it for the built-in server when `PDF_CONVERTER_AUTH=google-iam`, and the PDF page renderer client now reuses it.
- New workers sweep (`pdf-exports-sweep` queue) deletes objects under the tmp prefix older than TTL + 60 s every 5 minutes, with strict prefix validation.

### Web (`apps/web`, `apps/web-embed`, `apps/help`)
- The MCP App sandbox gains `allow-popups allow-popups-to-escape-sandbox` so the card's Download anchor opens a new tab; the host also handles `ui/open-link` honestly (http(s) only, transient-activation and blocked-popup checks, opener severed).
- Built-in servers show a "Built-in" badge and no delete action in the server list and the agent MCP tab; i18n in en and fr.
- Stories: `McpServersList` (Empty / BuiltInOnly / WithServers), `AgentEditor` with the built-in, and a new `common/McpAppView` story that mounts the real double iframe with a sample card.
- Help pages updated in en and fr; changelog entries added.

## Verification
- Go: vet, tests, gofmt and `CGO_ENABLED=0` build all clean.
- Repo: `npm run typecheck` and `npm run biome:check` exit 0; `apps/api` `test:parallel` passed (304 suites, 2336 tests); `check:boundaries` clean; `vitest run mcp-app-view` 8/8.
- End to end: the converter was run locally against the dev bucket with a service-account key and driven with the API's own client library (`@ai-sdk/mcp`): tool listing, card resource read, conversion, upload, signed URL and download (200, `%PDF-`, `Content-Disposition` with `filename` and RFC 5987 `filename*`) all worked.
- Not yet exercised: the full stack in a browser (built-in synced at API boot, card rendered in an agent session, Download in a new tab, expired state, workers sweep log). Storybook `common/McpAppView/PdfExportCard` is the quick check for the sandbox flags.

## Infra follow-ups (out of this repo)
- Converter service account: `roles/iam.serviceAccountTokenCreator` on itself plus the IAM Credentials API, otherwise every export returns an error result.
- Workers service account: `storage.objects.list` and `storage.objects.delete` on the bucket; production `WORKER_QUEUE_NAMES` must include `pdf-exports-sweep`.
- A daily lifecycle rule on `tmp/` is a backstop only (lifecycle rules have day granularity).

## Notes for reviewers
- Overlaps with the in-flight `mcp-oauth` branch on `mcp-servers.service.ts`, the controller, the DTO and the module.
- Deliberate deviations from the plan: Unicode letters kept in file names (RFC 5987 header), built-in defined by an allowlist, boot sync non-fatal, card uses the native anchor rather than `ui/open-link`.

## Also in this PR: workers could not boot on `main`
Since da4691bf, `ExtractionAgentSessionRunLlmService` injects `ProjectsService`, but `ExtractionAgentSessionWorkersModule` never imported `ProjectsModule`, so the workers process failed at startup with a dependency-resolution error. Found while running the local stack to test this feature; fixed by adding the import (one file, separate commit 4f9d3753).

## Local end-to-end run (dev VM)
Converter, API, workers and web were started from this branch against the dev bucket: the API boot log shows the built-in server synced, the `mcp_server` row is created once and not rewritten on later restarts, the workers register the sweep and its first run completes against the real bucket, and the MCP smoke (tool list, card resource, conversion, signed download) passes. The browser session test is still to be done by hand.


alexisjamet:
Follow-up from the first browser test on the dev VM (commit 46368936):

- **Download did nothing**: the tab opened on the signed URL but never saved the file. Browsers judge a download by the frame that started the navigation, and the card lives in a sandboxed iframe, so both nested frames now carry `allow-downloads` (web and embed).
- **Card rendered dark on a light page**: it followed the OS colour scheme. It now follows `hostContext.theme` (light by default, the app has no dark mode) and the Download button is black on white.
- **Blank reply after a page refresh**: the stored message is complete and the API re-reads the card HTML, so the iframe handshake fails client-side after a refresh. Until the cause is found, a card that fails to render now logs `MCP App render failed` at warning level and the message shows the reply text instead of a blank.

PR #754, opened 2026-09-08
fix(web): return to the MCP OAuth callback after an Auth0 re-login

alexisjamet:
When the user comes back from an MCP provider's consent screen and Auth0 cannot silently restore the session (expired or revoked refresh token, cleared storage), `ProtectedRoute` redirects to Auth0 login without a `returnTo`. Auth0 sends the user back to the origin, `HomeRoute` navigates to onboarding, and the single-use `code` and `state` are lost with no error shown. The pending context also stays in localStorage for a later callback to pick up.

`ProtectedRoute` now passes `appState: { returnTo: pathname + search }` to `loginWithRedirect`, and `Auth0Provider` gets an `onRedirectCallback` that navigates back to it. Since the provider sits above `RouterProvider`, `Router.tsx` exposes the single data router instance and the callback uses `router.navigate(returnTo, { replace: true, flushSync: true })`, so the location commits before the SDK flips `isAuthenticated` and `HomeRoute` cannot race it to onboarding. A small `sanitizeReturnTo` helper accepts only same-origin relative paths (rejects `//host`, `/\host`, absolute URLs, control characters); anything else falls back to the SDK's default `history.replaceState`.

Testing: `npm run biome:check`, `npm run typecheck`, and `npx vitest run` in apps/web all pass (18 files, 152 tests, including 8 new tests for `buildReturnTo` and `sanitizeReturnTo`).

Addresses https://github.com/bayesimpact/bayes-platform/pull/710#discussion_r3915044385

### DIFF
--- a/apps/web/src/common/routes/ProtectedRoute.tsx
+++ b/apps/web/src/common/routes/ProtectedRoute.tsx
@@ -1,8 +1,10 @@
 import { useAuth0 } from "@auth0/auth0-react"
 import { useEffect } from "react"
+import { useLocation } from "react-router-dom"
 import { selectTermsAccepted } from "@/common/features/me/me.selectors"
 import { useAppSelector } from "@/common/store/hooks"
 import { AUTH0_ORGANIZATION_ID } from "@/config/auth0.config"
+import { buildReturnTo } from "./auth0-return-to"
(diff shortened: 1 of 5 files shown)


alexisjamet:
Closing without merging: decided not to pursue this change.

PR #760, opened 2026-09-08
fix(pdf-export): show the download card in the agent's language

alexisjamet:
## Summary

The built-in PDF export MCP server renders an MCP App card (file name, page count, size, countdown, Download button) that was English only. The card now follows the agent's configured language: French when the agent locale is `fr`, English otherwise.

## How it works

- **API**: the conversation context the API already forwards as headers on every MCP call (`X-Bayes-Agent-Id`, `X-Bayes-Session-Id`, visitor id) gains the agent's language as the standard `Accept-Language` header. Tool calls send the locale of the settings revision being run; the card resource reads (conversation and public chat) send the agent's current published locale. Documented in `apps/api/src/external/mcp/README.md`.
- **Converter** (`apps/pdf-converter/mcp.go`): `resources/read` picks the language from the first `Accept-Language` tag (`fr` or `fr-*` gives French, anything else English) and stamps it on the card's `<html lang>`.
- **Card** (`apps/pdf-converter/card.html`): starts from that `lang` attribute and renders every visible string from an en/fr dictionary, keeping its current state (preparing, result, error, cancelled) so a language change redraws in place. Sizes follow the language too (`2.3 MB` vs `2,3 Mo`). A `hostContext.locale` sent by an MCP Apps host still overrides it, though the platform host does not send one.

No DTO or web change: the agent DTO and the chat surfaces are untouched.

## Tests

- Go: `resources/read` with `Accept-Language: fr`, `fr-CA,fr;q=0.9,en;q=0.8`, `en`, `de` and no header returns the expected `lang`; a table test covers the tag parsing (run in the `golang:1.25-bookworm` image, since no local Go toolchain).
- API: header builder forwards `Accept-Language` and omits it when unknown; the tools service and the card HTML service pass the locale in the MCP context; public chat resolves it from the published settings. Affected unit and e2e suites pass (41 tests).
- Card protocol exercised end to end with a stubbed DOM harness, booting from `lang="en"` and `lang="fr"`, through tool-input, tool-result, host-context change, expired link, error and cancelled states.
- `npm run typecheck` and `npm run biome:check` pass.

### DIFF
--- a/apps/web/src/common/features/agents/external/agents.api.ts
+++ b/apps/web/src/common/features/agents/external/agents.api.ts
@@ -46,6 +46,7 @@ const toAgent = (dto: AgentDto): Agent => ({
   name: dto.name,
   projectId: dto.projectId,
   type: dto.type,
+  locale: dto.locale,
   currentRevision: {
     updatedAt: dto.currentRevision.updatedAt,
     name: dto.currentRevision.name,
--- a/apps/web/src/common/features/agents/external/agents.api.ts
+++ b/apps/web/src/common/features/agents/external/agents.api.ts
@@ -46,7 +46,6 @@ const toAgent = (dto: AgentDto): Agent => ({
   name: dto.name,
   projectId: dto.projectId,
   type: dto.type,
-  locale: dto.locale,
   currentRevision: {
     updatedAt: dto.currentRevision.updatedAt,
     name: dto.currentRevision.name,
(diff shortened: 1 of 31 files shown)


apps/web/src/common/features/agents/external/agents.api.ts
```diff
@@ -46,6 +46,7 @@ const toAgent = (dto: AgentDto): Agent => ({
   name: dto.name,
   projectId: dto.projectId,
   type: dto.type,
+  locale: dto.locale,
```
alexisjamet: nope

PR #762, opened 2026-09-09
feat(web): runtime configuration and an app-runtime image (front + API)

jdoucy:
## Why

The web front reads its settings from `VITE_*` variables that Vite inlines at build time. Each deployment needs its own build of the front. A Kubernetes chart must configure the front at deploy time, with values and environment variables, the way Grafana, Keycloak or Argo CD do. This PR makes that possible and ships a Docker target that serves the front and the API from one image.

## What changes

**apps/web**
- `src/config/runtime-config.ts`: one module reads `window.__CONFIG__` first, then the `VITE_*` values as a fallback for `vite dev` and static hosting builds. All 13 former `import.meta.env.VITE_*` reads go through it.
- Auth0 redirect and logout URLs use `getAppUrl()`, `public/` assets use `publicAssetUrl()`: the SPA works at `/` (static hosting) and at `/app` (served by the API). The router gets its `basename` from the Vite base path.
- `vite.config.ts`: `base` comes from `VITE_BASE_PATH` (default `/`).

**apps/api**
- `common/web-app`: when `WEB_APP_DIST_DIR` points to a built front, the API serves it under `/app`. `index.html` gets `window.__CONFIG__` injected from `WEB_*` variables, with defaults for the same origin and the Auth0 tenant the API already validates. Hashed assets are immutable, the entry page is never cached. `/` redirects to `/app/`. Nothing is registered when the variable is unset.
- The API routes stay at the root. The SPA lives under `/app` so a browser reload on a client route never hits an API route.

**Dockerfile**
- New `web-build` stage and `app-runtime` target (`FROM api-runtime` plus the `dist`). The `api-runtime`, `cpu-workers-runtime` and `gpu-workers-runtime` targets do not change.

## Checks

- `npm run biome:check`, `npm run typecheck`: pass.
- Web: 165 vitest tests pass, including the new `runtime-config.spec.ts`.
- API: 16 new jest tests pass (`src/common/web-app`). The full API suite runs in CI (no local test database in this worktree).
- `docker build --target app-runtime`: builds. The image contains `apps/web/dist` with `/app/` asset paths, sets `WEB_APP_DIST_DIR`, and adds 10 MB to `api-runtime`. `react` and `react-dom` were already present in `api-runtime`; this PR adds no front package to the production `node_modules`.

## Not in this PR

- No change to the production deployment of the front.
- No `/api` prefix on the API routes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 
 ### Added
 
+- Self-hosting: one Docker image can serve the web app and the API together, configured by environment variables.
 - (beta) Priority calls: agents on Gemini 3.x models can now enable the priority tier from the Model tab.
 - Agents can run on gemini-3.7-flash and gemini-3.8-flash.
 - Public agent banner: an optional notice pinned above the conversation on the public page and in the widget.
--- a/apps/api/.env-example
+++ b/apps/api/.env-example
@@ -131,3 +131,18 @@ VLLM_xxx_API="<not used>"
 # PDF_EXPORT_TMP_PREFIX=tmp/pdf-exports/
 # Required by every worker process (no default in code); must match the name listed in WORKER_QUEUE_NAMES.
 PDF_EXPORTS_SWEEP_QUEUE_NAME=pdf-exports-sweep
+
+# Web front served by the API (app-runtime image only). Unset locally: `vite dev` serves the front.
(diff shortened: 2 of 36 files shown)


jdoucy:
The trivy failure is not from this branch: the lockfile is unchanged here, the same findings (multer 2.2.0, @xmldom/xmldom 0.9.10 and 0.8.13) fail on the workers images too, and the fix is bayesimpact/bayes-platform#763. Merge #763 first, then update this branch from main and the security check passes.

Issue #765, opened 2026-09-09
Public API: freeze the public contract behind explicit validation and publish a versioned documentation


jdoucy:
## Context

External applications call the public API today. They use the routes under the `public` namespace (`packages/api-contracts/src/public-chat/`, ADR 0015). Those routes are a contract. A change to a path, a payload, or a response shape breaks the external applications without warning.

Today the contract has no public documentation and no version. `docs/public-chat-api.md` describes the routes, but it lives in the repository, it is not published, and nothing keeps it in sync with the code. #735 covers the OpenAPI document and the SDK. This issue covers the rule and the documentation that must exist before that work.

## Goal

1. Protect the public contract. Nobody changes it without an explicit validation.
2. Publish a clean, versioned documentation of the public API for external integrators.

## Part 1: rule "do not change the public contract without explicit validation"

Add the rule to `AGENTS.md` (root) and to `apps/api/AGENTS.md`:

- The public contract is every route, DTO, header, error code, and SSE event under `PUBLIC_PATH_PREFIX`.
- Do not change the public contract without an explicit validation from the maintainers. Ask first. Wait for a written "go" on the issue or the PR.
- Add a field, a route, or an optional property: allowed, with validation, as a minor version.
- Rename, remove, or change the type of an existing element: breaking change. It needs a new major version and a deprecation period for the old version.
- Every change to the contract updates the public documentation and the changelog in the same PR.

Enforce the rule in CI:

- A workflow detects a diff in `packages/api-contracts/src/public-chat/**`, in the public controllers, and in the public documentation.
- The check fails unless the PR carries a `public-contract-approved` label set by a maintainer.
- The check also fails when the contract changes and the documentation version does not.

## Part 2: public, versioned documentation

- Publish the documentation in `apps/help` (the public help site). Move and rewrite the content of `docs/public-chat-api.md`. Keep `docs/public-chat-api.md` as a pointer to the published page.
- The documentation covers: authentication (embed token, session token header), each route with request and response examples, the SSE event catalog, the error codes, the rate limits if any, and the CORS behavior (ADR 0015).
- Give the contract a version. Show it in the documentation, in a response header, and in the changelog. Start at `v1`.
- Add a public changelog page. Each entry gives the version, the date, the change, and the migration steps when the change is breaking.
- Define the deprecation policy in the documentation: how long an old version stays available, and how integrators get the notice.
- Write the documentation in Simplified Technical English: short sentences, one instruction per sentence, active voice.

[...]

PR #766, opened 2026-09-09
feat(public-api): version the public chat API contract under /public/v1 and freeze it

alexisjamet:
## Summary

Part 1 of #765 (code). Merge first, then #767 (docs), then #768 (gate). Infra companion: bayesimpact/infra#217.

- Routes move to `/public/v1/agents/{embedToken}/...`. The unprefixed `/public/agents/...` paths stay served as a plain alias of v1 (`PublicChatLegacyRoutes`, same routes, same bodies, same errors), so the deployed widget, launcher, public page and any client built on the released unprefixed API keep working, `bannerText` and MCP App cards included. The v1 controller serves both path sets from the same methods.
- One server folder per served major: `apps/api/src/domains/public-chat/v1/` with its controller, DTO mappers and e2e tests. `docs/public-api-contract.md` carries the checklist for shipping a v2.
- `PUBLIC_API_VERSION = "1.0"` and `PUBLIC_API_MAJOR = "v1"` in api-contracts. The major is a literal typed from the version string, so bumping to `2.0` fails typecheck instead of moving every route. No version header: the major is in the path, minors live on the changelog page.
- Public DTOs become self-contained and pinned to the shared types with compile-time `PublicContractPins`: a change to `StreamEventPayload`, `AgentSessionToolCallDto`, `AgentSessionMcpAppDto` or `AgentSessionMcpAppHtmlDto` now fails `typecheck` inside the contract folder.
- Contract description fixes: the stream route is declared GET as served, `EmbedPublicConfigDto` moves to `public-chat.dto.ts`, the stream response is typed.
- Cleanup from the issue comment: DTO mapping leaves the service for the per-version mappers, `createSession` returns `CreatePublicSessionResponseDto`, dead `toEmbedConfigDto` and an unused constructor dependency are removed. Every stream failure (bad payload, missing agent, failed generation) is one JSON `{ "type": "error" }` event on a 200 response, then the end of the stream. No more bare `event: error` frame, which the reference client silently dropped, and no dependence on an unrelated interceptor.
- New `v1/e2e-tests/public-contract.spec.ts` pins the literal paths, error bodies, SSE framing and CORS. `v1/e2e-tests/legacy-alias.spec.ts` pins that the alias declares every v1 route and answers with the same bodies and the same errors.
- The rule lives once in `docs/public-api-contract.md`; root `AGENTS.md`, `CLAUDE.md`, `apps/api/CLAUDE.md` and a new `apps/api/AGENTS.md` point to it. `docs/public-chat-api.md` becomes a pointer to the help pages.
- web-embed builds its URLs from the route helpers and moves to v1.

## Maintainers

This PR changes the public contract. Under the new rule it needs a written go and the `public-contract-approved` label from a maintainer other than the author (@jdoucy, @did or @OlivierDoucy).

## Verification

`biome:check`, `typecheck`, `check:boundaries`, the web suite, the web-embed build and the full API suite (320 suites, 2664 tests) pass. Adding a field to `AgentSessionToolCallDto` fails typecheck at the pin.


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ## [Unreleased]
 
 ### Added
+- Public chat API: the routes are now versioned under /public/v1/ and documented on the help site.
```
jdoucy, line 11: merge the 3 k8s into one, just saying, adding chart

PR #768, opened 2026-09-09
ci: gate public API contract changes behind a maintainer label

alexisjamet:
## Summary

Part 3 of #765 (enforcement). Merge last: the script reads the help pages' `apiVersion`, so it needs the code and documentation PRs on main.

- New `public-contract` workflow on `pull_request` (including `labeled` and `unlabeled`), always reporting, so it can become a required check.
- `.github/scripts/public-contract-check.mjs` fails the PR when contract files or the public help pages change without `public-contract-approved`, when the label was set by a non-maintainer or by the PR author, when `packages/api-contracts/src/public-chat/` changes without a `PUBLIC_API_VERSION` bump or a changelog entry, or when a help page's `apiVersion` differs from the constant.
- `.github/CODEOWNERS` auto-requests the four maintainers on the contract paths.

This PR gates itself: the check is red until a maintainer other than the author sets the label.

## After merge (admin)

Add `public-contract` to the required status checks of ruleset 14610525 (read-modify-write on `repos/bayesimpact/bayes-platform/rulesets/14610525`, or Settings, Rules, main). Consider requiring `ci-checks` too, which is what enforces the type pins.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,8 +4,6 @@
 # review automatically and names them.
 /packages/api-contracts/src/public-chat/                            @jdoucy @alexisjamet @did @OlivierDoucy
 /apps/api/src/domains/public-chat/v1/                               @jdoucy @alexisjamet @did @OlivierDoucy
-/apps/api/src/domains/public-chat/legacy/                           @jdoucy @alexisjamet @did @OlivierDoucy
-/apps/api/src/domains/public-chat/public-chat.controller.ts         @jdoucy @alexisjamet @did @OlivierDoucy
 /apps/api/src/domains/public-chat/guards/                           @jdoucy @alexisjamet @did @OlivierDoucy
 /apps/api/src/config/cors.ts                                        @jdoucy @alexisjamet @did @OlivierDoucy
 /apps/help/src/content/docs/en/public-chat-api*.mdx                 @jdoucy @alexisjamet @did @OlivierDoucy
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,17 @@
+# Public API contract (docs/public-api-contract.md). A change needs a written go from one of
+# these maintainers and the public-contract-approved label, set by a maintainer who is not
+# the PR author. The ruleset does not require code owner review; this file requests their
+# review automatically and names them.
+/packages/api-contracts/src/public-chat/                            @jdoucy @alexisjamet @did @OlivierDoucy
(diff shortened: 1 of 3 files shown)


alexisjamet:
## Follow-up from #766 (commit 456cbfa1): the API folder is now split per served contract

The code PR now has one server folder per contract, `apps/api/src/domains/public-chat/v1/` and `legacy/`. Three of the paths this PR watches no longer exist at their old location, so the gate would let a change to the actual controllers, mappers or contract spec through without the label.

Paths to update in both `.github/scripts/public-contract-check.mjs` (`LABEL_PATHS`) and `.github/CODEOWNERS`:

| Current entry | Now |
| --- | --- |
| `apps/api/src/domains/public-chat/public-chat.mappers.ts` | gone. Mappers live in `v1/public-chat-v1.mappers.ts` and `legacy/public-chat-legacy.mappers.ts` |
| `apps/api/src/domains/public-chat/e2e-tests/public-contract.spec.ts` | moved to `v1/e2e-tests/public-contract.spec.ts`, plus new `legacy/e2e-tests/legacy-alias.spec.ts` |
| `apps/api/src/domains/public-chat/public-chat.controller.ts` | still exists but is only the abstract base. The route bodies are in `v1/public-chat-v1.controller.ts` and `legacy/public-chat-legacy.controller.ts` |

Suggested replacement, matching the "Contract files" list in `docs/public-api-contract.md` on #766:

```
apps/api/src/domains/public-chat/v1/
apps/api/src/domains/public-chat/legacy/
apps/api/src/domains/public-chat/public-chat.controller.ts
apps/api/src/domains/public-chat/guards/
```

Watching the two folders whole also covers the per-version e2e tests, which is intended: a change to `v1/e2e-tests/` is a change to what v1 is pinned to serve. A future `v2/` folder will need its own entry, which the checklist in `docs/public-api-contract.md` can mention.

`VERSION_PATHS` needs no change: `LegacyEmbedPublicConfigDto` and `PublicChatLegacyRoutes` sit in `packages/api-contracts/src/public-chat/`, and the alias is frozen, so any edit there is a contract change by definition.

alexisjamet:
Applied in 8fb317c8: the gate script and CODEOWNERS now watch `apps/api/src/domains/public-chat/v1/` and `legacy/` in place of the pre-split mappers file and contract spec path. `node --check` passes.

alexisjamet:
Checked against the final layout of #766 after 9f4d4ff3: the watched paths in the script and CODEOWNERS are correct as they stand. `v1/`, `legacy/`, `public-chat.controller.ts` and `guards/` all exist, and the `legacy/` entry matters, since that folder carries the frozen contract of the unprefixed paths. It goes away with the folder once the last integrator migrates.

PR #771, opened 2026-09-09
feat(dev): start the pdf-converter service with npm run dev

alexisjamet:
## Summary

`npm run dev` now also starts the Go pdf-converter service, so `http://localhost:3002/mcp` (the PDF export MCP endpoint) is up alongside the API and the web app.

- `apps/pdf-converter` becomes a workspace package with a one-line shell `dev` script that Turbo runs like the other apps. It exits 0 with a message when `go` is not on `PATH` or when `apps/pdf-converter/.env` is missing, so the other apps still start. Otherwise it runs `go run .`.
- The service owns its config: new `apps/pdf-converter/.env-example` with `GCS_STORAGE_BUCKET_NAME`, `GOOGLE_APPLICATION_CREDENTIALS` and an optional `PORT`. `main.go` loads `.env` with `github.com/joho/godotenv`, a no-op when the file is absent, so the Dockerfile and production are unchanged. Variables already set in the environment win.
- The missing-bucket error now points to `.env-example`.
- Docs: one line in the root README and CLAUDE.md, a short "Local development" note in the service README. Lockfile only gains the two workspace entries.

No CHANGELOG entry: developer tooling only, no end-user capability.

## Test plan

- [x] `go build`, `go vet`, `go test ./...` and `go mod verify` pass in `apps/pdf-converter`
- [x] With `.env` present, `npm run dev` starts the service, `GET /healthz` returns 200, `POST /mcp tools/list` returns `convert_markdown_to_pdf`
- [x] Without `go` or without `.env`, the script prints the skip message and exits 0
- [x] `turbo run dev --dry-run` lists `@caseai-connect/pdf-converter#dev`
- [x] `npm run biome:check`, `npm run typecheck`, `npm ci --dry-run` pass


jdoucy (changes requested):
Thanks for wiring this into `npm run dev`, that is the right goal. I would like a simpler shape before merging though.

Turbo runs whatever `package.json` script it finds, so it can call `go run .` directly. The 74-line Node launcher exists to work around three things, and each has a cheaper answer:

**Looking for Go outside PATH** (GOROOT, /usr/local/go/bin, ~/go/bin). If `go` is not on PATH the developer's setup is broken, and the repo should not paper over it. `command -v go` is enough.

**Reading `apps/api/.env` from another app.** This couples the Go service to the API's config file. Every other app owns its `.env`. The Go service should do the same: add `apps/pdf-converter/.env-example` with `GCS_STORAGE_BUCKET_NAME` and `GOOGLE_APPLICATION_CREDENTIALS`, and load it in `main.go` with `godotenv.Load()` (a no-op when the file is missing, so the Dockerfile and prod are untouched). Root `.gitignore` already ignores `.env`.

**Signal forwarding.** Turbo already forwards SIGINT/SIGTERM to persistent tasks, the wrapper adds nothing here.

Suggested `apps/pdf-converter/package.json`:

```json
{
  "name": "@caseai-connect/pdf-converter",
  "private": true,
  "version": "0.0.0",
  "scripts": {
    "dev": "command -v go >/dev/null || { echo '[pdf-converter] go not found, skipping'; exit 0; }; PORT=${PDF_CONVERTER_PORT:-3002} go run ."
  }
}
```

This drops `scripts/dev.mjs`, most of the added docs, and the `apps/api/.env` dependency.

One case this does not cover: exiting 0 when no bucket is configured. I would rather have `main.go` print a clear message and exit than keep a launcher for that. Someone who does not want the service can also run `turbo run dev --filter=!@caseai-connect/pdf-converter`.

Docs: keep the one-line mention in the root README and CLAUDE.md, and reduce the service README addition to "copy `.env-example` to `.env`, then `npm run dev` at the root starts it".

alexisjamet:
Refactored as suggested in f1390061 and 171aa591.

- `scripts/dev.mjs` is gone. `apps/pdf-converter/package.json` now has a one-line shell `dev` script: skip with exit 0 if `go` is not on `PATH`, otherwise `go run .`.
- The service owns its config: new `apps/pdf-converter/.env-example` with `GCS_STORAGE_BUCKET_NAME`, `GOOGLE_APPLICATION_CREDENTIALS` and an optional `PORT`. `main.go` loads `.env` before reading the environment, the file is absent in the image and in prod so nothing changes there. No more dependency on `apps/api/.env`.
- Docs reduced to the one-line mentions in the root README and CLAUDE.md, plus "copy `.env-example` to `.env`, then `npm run dev`" in the service README.
- `main.go` now prints a clear message pointing at `.env-example` when the bucket is missing.

Two deviations from your snippet:

- The script also skips with exit 0 when `apps/pdf-converter/.env` does not exist. Turbo 2.9 defaults to `--continue=never`, so a non-zero exit from the Go service would cancel the API and web tasks too. Keeping `main.go` strict and letting the shell skip when the file is absent covers the fresh-clone case without a launcher.
- `.env` is read by a 50-line in-house `dotenv.go` (blank lines, `#` comments, `export` prefix, quoted values, existing env wins, missing file is a no-op) with tests, instead of `godotenv`. We preferred not to add a third-party dependency for a file reader. `PDF_CONVERTER_PORT` is dropped since `PORT` can live in the service's `.env` and defaults to 3002.

jdoucy (commented):
I prefer this stable lib:
https://github.com/joho/godotenv

alexisjamet:
Back to `github.com/joho/godotenv` in 4b0e22e9, which reverts the in-house loader commit. `main.go` calls `godotenv.Load()` again, `dotenv.go` and its test are gone, and go.mod/go.sum carry v1.5.1. `go build`, `go vet`, `go test ./...` and `go mod verify` pass. PR description refreshed to match the current shape.

jdoucy (approved):

PR #777, opened 2026-09-09
fix(web): show the identity provider error instead of looping back to Auth0

jdoucy:
Auth0 reports a failed authorization by redirecting to the app with `error` and `error_description` in the query string, for example `?error=invalid_request&error_description=parameter organization is not allowed for this client`. `HomeRoute` and `ProtectedRoute` called `loginWithRedirect` whenever the user was not authenticated, so the browser bounced between the app and Auth0 without end, with a new `state` at each turn.

## What changes

- `common/auth/auth-callback-error.ts`: reads the error from the query string (unit tested).
- `AuthErrorRoute`: shows the provider's message, a "try again" that drops the error parameters and lets the home route start a clean login, and a "log out". Translated (en, fr).
- `HomeRoute` and `ProtectedRoute` render it when the callback carries an error, and skip the redirect.

## Checks

`npm run biome:check`, web typecheck, web tests (vitest, all green, 3 new).

### DIFF
--- a/apps/web/src/common/routes/HomeRoute.tsx
+++ b/apps/web/src/common/routes/HomeRoute.tsx
@@ -1,7 +1,7 @@
 import { useAuth0 } from "@auth0/auth0-react"
 import { useEffect } from "react"
 import { useNavigate, useSearchParams } from "react-router-dom"
-import { getAuthCallbackError } from "@/common/auth/auth-callback-error"
+import { useAuthCallbackError } from "@/common/auth/use-auth-callback-error"
 import { selectOrganizationsData } from "@/common/features/organizations/organizations.selectors"
 import { ADS } from "@/common/store/async-data-status"
 import { useAppSelector } from "@/common/store/hooks"
@@ -36,7 +36,7 @@ export function HomeRoute() {
   const organizations = useAppSelector(selectOrganizationsData)
   const [searchParams] = useSearchParams()
   // Auth0 sent us back with an error: show it instead of redirecting again.
-  const callbackError = getAuthCallbackError(searchParams.toString())
+  const callbackError = useAuthCallbackError()
 
   useEffect(() => {
     if (isLoading || callbackError) return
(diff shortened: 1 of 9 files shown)


apps/web/src/common/routes/ProtectedRoute.tsx
```diff
@@ -1,17 +1,24 @@
...
 import { LoadingRoute } from "./LoadingRoute"
 import { TermsRoute } from "./TermsRoute"
 
 export function ProtectedRoute({ children }: { children: React.ReactNode }) {
   const { isAuthenticated, isLoading, loginWithRedirect } = useAuth0()
   const isPlatformLoading = useAppSelector((state) => state.auth.isLoading)
   const termsAccepted = useAppSelector(selectTermsAccepted)
+  const location = useLocation()
+  // Auth0 sent us back with an error: show it instead of redirecting again.
+  const callbackError = getAuthCallbackError(location.search)
```
jdoucy: group in a specific hook?

apps/web/src/common/routes/HomeRoute.tsx
```diff
@@ -33,9 +35,11 @@ export function HomeRoute() {
   const { isLoading, isAuthenticated, loginWithRedirect, logout } = useAuth0()
   const organizations = useAppSelector(selectOrganizationsData)
   const [searchParams] = useSearchParams()
+  // Auth0 sent us back with an error: show it instead of redirecting again.
+  const callbackError = getAuthCallbackError(searchParams.toString())
```
jdoucy: idem hook?

apps/web/src/common/routes/ProtectedRoute.tsx
jdoucy: Done: `useAuthCallbackError()` in `common/auth/use-auth-callback-error.ts`, used by both routes. The pure `getAuthCallbackError(search)` stays for the unit test.

apps/web/src/common/routes/HomeRoute.tsx
jdoucy: Done: `useAuthCallbackError()` in `common/auth/use-auth-callback-error.ts`, used by both routes. The pure `getAuthCallbackError(search)` stays for the unit test.

PR #778, opened 2026-09-09
fix(api): the status listeners honour DATABASE_SSL

jdoucy:
Seen on a cluster with Cloud SQL in `ENCRYPTED_ONLY` mode after #773: the API connects (TypeORM honours `DATABASE_SSL`), but the `LISTEN/NOTIFY` clients of `PostgresStatusStreamService` open their own `pg` connections without TLS and fail every reconnect (`pg_hba.conf rejects connection ... no encryption`, 89 times in 3 minutes). Live status updates (SSE) never reach the browser.

- `config/database-ssl.ts`: the TLS options (`DATABASE_SSL`, `DATABASE_SSL_CA`) in one place, unit tested.
- `config/typeorm.ts` uses it (no behaviour change).
- `postgres-status-stream.service.ts`: the listener client uses it too.

`npm run biome:check`, API typecheck and the new tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ### Changed
 
 ### Fixed
+- Live status updates now work on databases that require TLS.
 - Sign-in: an error returned by the identity provider is shown with a retry button instead of reloading endlessly.
 - (beta) MCP App cards: the reply text now stays on screen next to the card instead of disappearing once the card shows.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ### Changed
 
 ### Fixed
+- Live status updates now work on databases that require TLS.
```
jdoucy: remove it, not relevent

jdoucy: Removed.

PR #780, opened 2026-09-09
feat(rbac): platform roles granted by an operator (command and initialAdmins), never at sign-in

### LINKED ISSUE #779: First administrator bootstrap: grant platform roles at login and from a command
## Problem

On a fresh database, nobody can create an organization. The global roles (`platform_staff`, `platform_superadmin`) are given by two migrations that read `ORGANIZATION_CREATOR_EMAIL_DOMAIN` and `BACKOFFICE_AUTHORIZED_EMAILS` and assign the roles to the users that **already exist** when the migration runs. On a new install the migrations run before any user has logged in, so the first administrator lands on "You are not allowed to create an organization" and the only way out is an SQL insert in `user_membership`.

## Target

Roles are granted by an operator, never derived from a sign-in:

1. **An administration command** in the image: `platform-role grant|revoke|list --email <email> [--role ...]`. `grant` pre-provisions the user when the person never signed in (placeholder identity, linked at first sign-in like an invited member). Runs with `kubectl exec`, a Cloud Run Job or a Helm Job.
2. **`initialAdmins` in the chart values**: a post-install Job grants `platform_superadmin` to each listed email, after the migrations. Explicit, one-shot, reviewed in the deployment repository.
3. **Delegation in the back office**: a superadmin grants or revokes the global roles of a user from the user page. Later, with the identity provider work.

## Rejected: granting at sign-in

A first version granted the roles at sign-in from `BACKOFFICE_AUTHORIZED_EMAILS` and `ORGANIZATION_CREATOR_EMAIL_DOMAIN`. Rejected: the API would trust the email reported by the identity provider, and anyone able to register an address of the right shape (the domain, or a listed email on another connection) would become staff or superadmin without any operator action. The seed migrations apply the same rules once, at a moment chosen by the operator; making them continuous changes the nature of the risk. Also rejected: a "first user becomes admin" setup screen, a race behind a public URL.

## Scope of this issue

Layers 1 and 2: #780.

## Seen on

The first deployment of the Helm chart on a fresh database (2026-09-09): migrations ran, the first user logged in through Auth0, no global role, onboarding refused.

jdoucy:
Closes #779.

On a fresh database nobody can create an organization: the global roles (`platform_staff`, `platform_superadmin`) are only given by two seed migrations that reach the users that exist when they run. On a new install nobody exists yet, so the first administrator lands on "You are not allowed to create an organization".

## Decision

No automatic grant at sign-in. A role derived from the email the identity provider reports would let anyone able to register an address of the right shape (domain, or a listed email on another connection) become staff. Roles are granted by an operator.

## What changes

- **`PlatformRoleService`** (`domains/rbac`): `grantGlobalRole`, `revokeGlobalRole`, `listGlobalRoles`. Grant is idempotent (`INSERT ... WHERE NOT EXISTS`). Revoke is a hard delete, like `UserMembershipRepository.deleteMembership`: the unique index on (user, role) for global memberships ignores `deleted_at`.
- **`scripts/platform-role.ts`**: `grant | revoke | list --email <email> [--role platform_superadmin|platform_staff]`. `grant` works before the person has ever signed in: the user is created with a placeholder identity and linked to the real one at first sign-in, the same path as an invited member. `npm run platform-role -- ...` from the sources, `node /app/apps/api/dist/scripts/platform-role.js ...` from the runtime image.
- **Chart**: `initialAdmins` (list of emails) and a post-install, post-upgrade Job (weight 10, after the migrations) that grants `platform_superadmin` to each of them. Empty by default: no Job.
- Documentation: root README, chart README ("First administrator").

## Checks

- `PlatformRoleService` tests on the test database (grant once, list sorted, revoke then grant again, no-op on a user without roles) and the role key guard.
- `npm run biome:check`, API typecheck, `npm run check:boundaries`: pass. `helm lint` and `helm template` with two `initialAdmins`: the Job runs the command once per email.

## Not in this PR

Granting the roles from the back office (layer 3 of #779), with the identity provider work.


apps/api/src/domains/rbac/platform-role.service.ts
```diff
@@ -0,0 +1,95 @@
+import { Injectable } from "@nestjs/common"
```
jdoucy, line 1: we should have a clean repository, implementing sql stuff and service dontknow about typorm

jdoucy: Done: the SQL is now in PlatformRoleRepository (TransactionService.getManager(), same pattern as UserMembershipRepository). PlatformRoleService only delegates and has no TypeORM import. Repository spec runs against the database, service spec uses a mock.

PR #784, opened 2026-09-10
ci: workers smoke on the self-hosted runner

jdoucy:
## Summary

The workers smoke builds the CPU and GPU images with `docker build` and starts them with compose, 4 to 5 minutes on a GitHub runner, two of them installing Torch. On the self-hosted runner the Docker daemon keeps its build cache between runs, so only the layers that depend on the branch are rebuilt. Expected: 1 to 2 minutes.

- `runs-on: self-hosted`.
- Concurrency group: one smoke at a time, the compose stack binds fixed ports (55432, 56379, 3003, distinct from the tests).

Push events fire for branches of this repository only, never for forks, so no external code reaches the runner through this workflow.

The runner keeps a cap on the daemon build cache so the Torch layers are not evicted between two runs (deployment repository).

### DIFF
--- a/.github/workflows/branch-workers-smoke.yml
+++ b/.github/workflows/branch-workers-smoke.yml
@@ -12,7 +12,15 @@ on:
 
 jobs:
   workers-smoke:
-    runs-on: ubuntu-latest
+    # Self-hosted: the Docker daemon of the runner keeps the build cache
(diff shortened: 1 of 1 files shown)


jdoucy:
Replaced by #787, one pull request for the two changes plus the tests gate.

PR #785, opened 2026-09-10
fix(chart): MCP OAuth callback derived from the web URL

jdoucy:
## Summary

The chart did not set `MCP_OAUTH_REDIRECT_URL`. The API reads it with `getOrThrow` when a user authorizes an MCP server with OAuth, so every install answered a 500 on that action.

- Default in the shared env: `<urls.web>/oauth/mcp/callback`, the callback route of the web front. `config.MCP_OAUTH_REDIRECT_URL` overrides it.
- Values comment and changelog.

Checked the other settings the API reads with `getOrThrow` (Auth0 issuer, audience, organization, client ids, M2M secret, database connection name): all covered by the chart.

`helm lint` passes. `helm template` renders the default from `urls.web` and drops it when the override is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project uses [CalVer](https://calver.org/) (YY.MM.Micro) for product versio
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


jdoucy:
Replaced by #787, one pull request for the two changes plus the tests gate.

PR #799, opened 2026-09-15
fix(llm): strip the leaked tool call written as the bare tool name and its arguments

jdoucy:
## Summary

A new tool-call leak variant was seen in production on Gemma: `mandatory_tool{categoryNames:[...],suggestedTitle:...}`. There is no `<`, no `call:` prefix and no `default_api:` namespace. Every existing pattern needs one of these markers, so the text reached the user on consecutive replies and the tool did not run.

- The cleanup pattern gets a generic `identifier{key:value,...}` alternative. It is tool-name agnostic on purpose: the model can hallucinate the name too, and the leak must be hidden anyway. The argument object must start with a `key:`, so prose braces (`function foo() {}`, `{a, b}`) never match.
- The streaming sanitizer holds the buffer back on an open `identifier{key:` until its closing `}`.
- `findLeakedToolCalls` extracts the tool name from this variant too, so recovery still runs the tool (or reports an undeclared name).
- The helper is renamed `LLMOutputSanitizer` (`sanitize`, `createStreamSanitizer`): it strips more than thought tokens.
- Changelog entry.

Follows #620, #668 and #731.

## Test plan

- [x] New specs: full-text cleanup for a declared and a hallucinated tool name, prose with braces, streaming cleanup split across 7-char deltas, tool-name extraction
- [x] New recovery spec: the name-only variant runs the tool with the recovered arguments
- [x] `npx jest` on llm-output-sanitizer and leaked-tool-call-recovery specs (25 passing)
- [x] biome check on apps/api/src/external/llm
- [x] Confirm after release that the variant no longer shows in agent replies


apps/api/src/external/llm/thought-tokens-helper.spec.ts
```diff
@@ -57,6 +63,24 @@ describe("ThoughtTokensHelper - hallucinated tool-call XML", () => {
     expect(cleaned).toBe("Bonjour ! Comment puis-je vous aider aujourd'hui ? ")
   })
 
+  it("removes the name-only variant when the tool name is declared", () => {
```
jdoucy: Maybe rename to sanitizer, because it's not only linked to thought tokens

apps/api/src/external/llm/ai-sdk-llm-provider-base.ts
```diff
@@ -282,6 +286,9 @@ export abstract class AISDKLLMProviderBase extends AISDKLLMToolsMgmt implements
     const baseModel = this.getLanguageModel(args)
     // Bound for use inside the stream TransformStream, where `this` is the
     // transformer, not the provider.
+    // Tool names declared for this call: the sanitizer needs them to spot the
+    // bare `tool_name{...}` leak variant, which carries no other marker.
+    const stripOptions = { toolNames: declaredToolNames(args.config) }
```
jdoucy: remove this options, the regexp match must be toolname agnostic, because toolname can be halucinated

apps/api/src/external/llm/thought-tokens-helper.spec.ts
jdoucy: Renamed: the file is now llm-output-sanitizer.ts and the class LLMOutputSanitizer (sanitize / createStreamSanitizer).

apps/api/src/external/llm/ai-sdk-llm-provider-base.ts
jdoucy: Removed. The bare variant is now matched as a generic identifier{key:value,...} shape, with no tool-name list. A hallucinated tool name is hidden too, and recovery reports it as undeclared. Prose braces without a key: (function foo() {}, {a, b}) still pass.

PR #804, opened 2026-09-17
feat(agents): forms belong to the conversation, one per agent with its settings revision

jdoucy:
## What

The answers the fillForm tool collects move from a `result` JSON column on each session table to a `conversation_form` table: one row per (session, agent), with the agent settings revision of the last write, a status (`in_progress`, `concluded`) and the state. A conversation keeps every form filled in it. Design in `docs/adr/0017-conversation-forms.md`.

No visible change for users. This is the data model the next steps need: a sub-agent that takes the conversation over inside the parent's session, one form per agent in that session, the parent reading every form, a sub-agent seeing what earlier forms already know.

## Changes

- `ConversationForm` entity, `ConversationFormsService` (list, find, merge field by field, delete for a session), factory, module. `session_id` has no foreign key, like `agent_message.session_id`, so conversation and public sessions both own forms.
- fillForm tool writes to the form of `(session, agent)` and stamps `agent_settings_id`. `getFormState` reads the stored form instead of the session snapshot loaded at the start of the turn.
- The per-surface "result updater" is gone: public sessions write to the same table. The public streaming proxy carries `persistsForms` instead of a `result` snapshot; evaluation runs set it to false and keep no fillForm tool, as before.
- Session DTOs expose `forms: ConversationFormDto[]` instead of `result`. Studio and tester fronts pick the form of the agent they show (`findConversationForm`).
- Review campaigns: the reviewer reads a session's form with the settings revision that wrote it, so the answers are shown against the schema that collected them. The tester listing carries `forms`.
- Retention purge and session deletion drop the form rows (the answers are content, the row has no analytics value).
- Migration: schema generated from the entities, plus two `INSERT ... SELECT` that move the existing states, attributed to the session's agent and stamped with the revision of the session's last assistant message (or the agent's latest published revision). `down` puts the states back on the session tables. Checked up and down on a throwaway database with seeded sessions (Studio and embed).
- Boundary baselines regenerated for the new entity relation (`conversation_form` ↔ `conversation_agent_session`).

## Tests

- `conversation-forms.service.spec.ts`: first write creates, later writes merge, revision of the last write, one form per agent per session, connect scope.
- Updated: fillForm tool specs, public stream spec, sub-sessions listing, purge, reviewer and tester e2e.
- `npm run typecheck`, `biome ci`, `check:boundaries`, web vitest, API `test:parallel` green.

## Note for the gate

The public contract gate will ask for the `public-contract-approved` label: the change touches `public-chat/v1/e2e-tests/stream-messages.spec.ts` (assertion moved from the session column to the form table). The public routes and DTOs are unchanged.


apps/api/src/domains/agents/base-agent-sessions/base-agent-sessions.service.ts
```diff
@@ -31,6 +32,7 @@ export class BaseAgentSessionsService {
   }): Promise<void> {
     await this.dataSource.transaction(async (entityManager) => {
       await this.deleteSessionMessages({ entityManager, sessionId: agentSession.id })
+      await ConversationFormsService.deleteForSession(entityManager, agentSession.id)
```
jdoucy: why a static method, its not the common pattern?

apps/api/src/domains/agents/conversation-agent-sessions/retention/conversation-agent-session-purge.service.ts
```diff
@@ -41,11 +42,13 @@ export class ConversationAgentSessionPurgeService {
       if (!session || session.purgedAt) return false
 
       deletedDocumentFiles.push(...(await this.purgeSessionMessages(entityManager, sessionId)))
+      // The forms hold answers only: the rows go.
+      await ConversationFormsService.deleteForSession(entityManager, sessionId)
```
jdoucy: the same here, why this static method pattern?

apps/api/src/domains/review-campaigns/reviewer/reviewer.service.ts
```diff
@@ -175,37 +178,49 @@ export class ReviewerService {
...
-      }),
-      this.reviewRepository.findOne({ where: { sessionId, reviewerUserId } }),
-      this.reviewRepository.find({ where: { sessionId } }),
-      this.testerFeedbackRepository.findOne({ where: { sessionId } }),
-      this.conversationSessionRepository.findOne({
-        where: { id: sessionId },
-        select: { id: true, createdAt: true, result: true },
-      }),
-    ])
+    const [transcript, callerReview, allReviews, testerFeedback, session, form] = await Promise.all(
```
jdoucy: need to create another issue, but we need to avoid this kind of parallel data loading versus standard SQL joins

apps/api/src/domains/agents/base-agent-sessions/base-agent-sessions.service.ts
jdoucy: Agreed, it was a wiring shortcut, not a pattern. Now an instance method of ConversationFormsService that takes the caller's EntityManager (ADR 0007 2.3), injected here; the session modules get ConversationFormsModule through the shared module helpers. 5906abad.

apps/api/src/domains/agents/conversation-agent-sessions/retention/conversation-agent-session-purge.service.ts
jdoucy: Same fix: injected ConversationFormsService, instance method. The purge spec builds the service with it. 5906abad.

apps/api/src/domains/review-campaigns/reviewer/reviewer.service.ts
jdoucy: Replaced the sixth query by a join on the session load (relations forms and their agentSettings) and picked the campaign agent's form from it. The five earlier parallel reads are unchanged here; the wider replacement of parallel loads by joins is tracked separately, with the other candidate services. 5906abad.

PR #810, opened 2026-09-17
fix(agents): one hand-over per turn, and the hand-over ends the parent's turn

jdoucy:
## What

Seen on a real orchestration with Gemma after #805: the parent handed the conversation to one questionnaire sub-agent, kept generating, handed it to a second one in the same turn, and wrote a second transition sentence with a leftover `</code>`. The conversation went to the last sub-agent called, not to the one announced, and the reply carried two contradictory sentences.

## How

- **One hand-over per turn.** The handoff tools of a turn share a state: the first call wins, later calls return `already_handed_off` with the name of the agent that took over and the instruction to write one sentence and stop. The active agent is not moved again.
- **The hand-over ends the turn.** New `terminalToolNames` on the LLM config, set to the handoff tools: once one ran, the tool loop allows one more generation for the closing sentence (none when the hand-over step already carried it) and stops, whatever else the model would call. Same mechanism as the fire-and-forget stop condition.
- **Empty wrappers.** When the sanitizer removes a verbalized tool call, it also drops the `<code>…</code>` or fenced block the model wrapped it in, which stayed behind as `</code>` in the reply.

## Tests

- `terminal-tools-stop-condition.spec.ts`: no terminal tool, sentence on the hand-over step, one extra generation then stop, stop even when the model calls other tools instead of writing.
- `stream-handoff.spec.ts`: the parent hands over to one sub-agent, writes its sentence and calls a second hand-over in the same generation: the first wins, the second is refused, no third parent generation, the first sub-agent speaks next.
- Sanitizer: empty `<code>` wrapper and fence dropped, a real code block kept.
- `typecheck`, `biome ci`, `check:boundaries`, API `test:parallel` green.


jdoucy:
Superseded by #812, which carries these two commits with the rest of the live-run fixes.

